Issue #432, opened 2014-12-16
Bidi problems in deliteful/ToggleButton


semion1956:
Text direction isn't correctly controlled for checked state of the button.

PR #433, opened 2014-12-16
Control textDir in deliteful/ToogleButton fix#432

semion1956:
1) Move bidi logic into refreshRendering()
2) Use applyTextDirection() for setting textDir in checkedLabel
3) Support for textDir in title
4) updates in the template - not bidi-specific but suggested because FF gets error ("not well-formed") for ampersands in html.


semion1956:
*Why are these changes in ToggleButton instead of Switch?*
Changes are specific for ToggleButton.  What the connection between ToggleButton and Switch? I don't see it.
*It should have a unit test.*
You didn't require unit tests for Button... Anyway, if we need bidi-specific tests for Button and ToogleButton, should I add them into existing files or placed into new files?

wkeese:
> Changes are specific for ToggleButton. What the connection between ToggleButton and Switch? I don't see it.

I thought ToggleButton extended Switch, but I guess I was confused.  OK, nevermind.

> You didn't require unit tests for Button... Anyway, if we need bidi-specific tests for Button and ToggleButton, should I add them into existing files or placed into new files?

I think our rule is for everything to have unit tests, @cjolif shouldn't have pushed the Button code (462ae09dc3a151b89be984c71b8db2f7bbfc94fc) without them.

About the tests, I don't have a strong opinion about whether they should be in tests/unit/Button.js or another file (like tests/unit/Button_bidi.js).  The tricky part in either case is turning on the has("bidi") flag for your tests only.  I guess you need to do something tricky with requirejs contexts like https://github.com/ibm-js/requirejs-dplugins/blob/master/tests/unit/i18n.js does.  Perhaps @clmath has advice about that.

PS: You have another "Merge remote-tracking branch 'upstream/master'" commit in this branch.  Please stop calling "git merge".  Generally once people start commenting on a PR, you don't try to merge changes from the master branch until the PR is accepted.

cjolif:
> About the tests, I don't have a strong opinion about whether they should be in tests/unit/Button.js or another file (like tests/unit/Button_bidi.js)

I think if the bidi code is in Button itself ideally it should be in the same test file (if possible because of the has flag). If the bidi code is in a separate class then the test should probably be in a separate file as well.

wkeese:
Well, I think you need to do that regardless of whether or not both tests are in the same file.  Because both tests are unit tests.   Right?

clmath:
Maybe, I don't know how many loader instances intern is using.

semion1956:
Thanks @clmath.
I added intern tests.

tests/unit/all.js
```diff
@@ -23,7 +23,9 @@ define([
 	"./ToasterMessage-insert-show-hide-remove",
 	"./Toggle",
 	"./Button",
+	"./Button-bidi",
```
cjolif: for easier access to all bidi tests (and consistency with the fact bidi code is in bidi dir) maybe it should be bidi/Button and bidi/ToggleButton here.

cjolif: We could then have a unit/bidi/all.js that contains all of them to launch bidi test independently or exclude them all more easily (and tests/unit/all would reference tests/unit/bidi/all).

semion1956:
Tests are moved into tests/unit/bidi.

cjolif:
Looks ok to me. @wkeese any further remarks?

tests/unit/bidi/Button.js
```diff
@@ -0,0 +1,133 @@
...
+					}
+				}
+			});
+		},
+		markup: {
+			beforeEach: function () {
+				container = document.createElement("div");
+				document.body.appendChild(container);
+				container.innerHTML = html;
+			},
```
wkeese: This isn't quite right.   You are losing the pointer to the old container `<div>` without first destroying it.

tests/unit/bidi/Button.js
```diff
@@ -0,0 +1,133 @@
...
+			"auto2": function () {
+				var dfd = this.async();
+				moduleRequire(["delite/register", "deliteful/Button"], dfd.callback(function (register) {
+					register.parse();
+					b4 = document.getElementById("b4");
+					b4.deliver();
+					assert.strictEqual("\u202aABC \u05d2\u05d1\u05d0\u202c", b4.labelNode.textContent, "auto2: wrong displayed value for 'label'");
+					assert.strictEqual("\u202aABC \u05d2\u05d1\u05d0\u202c", b4.title, "auto2: wrong value for 'title'");
+				}));
+			}
```
wkeese: There are a bunch of calls to `register.parse()` that should really be `register.parse(container)`, or something like that.   Only parse on the DOM that you've created for each test run.

tests/unit/bidi/Button.js
```diff
@@ -0,0 +1,133 @@
...
+		},
+		dynChanges: {
+			beforeEach: function () {
+				container = document.createElement("div");
+				document.body.appendChild(container);
+			},
```
wkeese: Likewise here too, you are losing the pointer to the old container `<div>` without first destroying it.   Note that this `beforeEach()` method will run before each test within the `dynChanges` section, but the `afterEach()` method at line 129, outside of the `dynChanges` section, will only run once before `dynChanges` starts, and then once after it ends (in other words, before `textDir` and after `title`).

My suggestion is that:

1. In startup() for the main test suite you create container
2. In teardown() for the main test you destroy container
3. For other places you just empty container, like `container.innerHTML = ""`

tests/unit/bidi/ToggleButton.js
```diff
@@ -0,0 +1,167 @@
...
+					b3.title = "ABC \u05d0\u05d1\u05d2";
+					b3.deliver();
+					assert.strictEqual("\u202aABC \u05d0\u05d1\u05d2\u202c", b3.title, "title: wrong value for 'auto' (2)");
+				}));
+			}
+		},
+		afterEach: function () {
+			container.parentNode.removeChild(container);
+		}
+	});
```
wkeese: Same comments here as for the Button test.

semion1956:
Done.

cjolif:
ref #432

ToggleButton/ToggleButton.html
```diff
@@ -1,4 +1,4 @@
 <template aria-pressed="{{checked}}" attach-point="focusNode" on-click="{{toggle}}">
-	<span attach-point="iconNode" class="d-reset d-inline d-icon {{this.checked && this.checkedIconClass ? this.checkedIconClass : (this.iconClass ? this.iconClass : 'd-hidden')}}"
-	></span><span attach-point="labelNode" class="d-reset d-inline" d-hidden="{{!this.checkedLabel && !this.label}}">{{this.checked && this.checkedLabel ? this.checkedLabel : this.label}}</span>
+	<span attach-point="iconNode" class="d-reset d-inline d-icon {{this.checked &amp;&amp; this.checkedIconClass ? this.checkedIconClass : (this.iconClass ? this.iconClass : 'd-hidden')}}"
+	></span><span attach-point="labelNode" class="d-reset d-inline" d-hidden="{{!this.checkedLabel &amp;&amp; !this.label}}">{{this.checked &amp;&amp; this.checkedLabel ? this.checkedLabel : this.label}}</span>
```
AdrianVasiliu: There might be a reason, but I wonder why you escape the `&&` into `&amp;&amp;`. 
https://github.com/ibm-js/delite/blob/master/docs/handlebars.md doesn't mention such a need, and at least in some browsers `&&` seems to work just fine.

AdrianVasiliu: Oops, nevermind, I didn't notice it has been already discussed (in https://github.com/ibm-js/deliteful/pull/433#commitcomment-9123284 ).

cjolif:
merged in 878fd8381a11d9b8bf9e155a8df1d8546cbbcc88. thanks @semion1956

PR #437, opened 2014-12-19
Removed isAriaListbox flag, fixes #210

### LINKED ISSUE #210: List: easy way to switch between selection mode and action mode
Switching lists between selection mode and action mode is a common UI paradigm.   Ex: click the "Edit" button on the list of messages in iOS's mail app.  Or the delete button on Android's mail app (when looking at your list of messages).

IIUC, it's possible to do this with List by switching `selectionMode` from `none` to `single` (or `multiple` or `radio`).   But it doesn't seem to mesh well with the new `isAriaListbox` flag.

What is the procedure for switching a list between selection mode and action mode when using that `isAriaListBox` flag?   Is it:

```js
myListBox.isAriaListBox = false;
myListBox.selectionMode = "single";
```

and then to switch back:

```js
myListBox.selectionMode = "none";
myListBox.isAriaListBox = true;
```

(And if so, does the order of the two commands matter?)

tkrugg:
- Removed isAriaListbox flag
- updated doc and samples

I set up a [mirror branch](https://travis-ci.org/tkrugg/deliteful/builds/44578571) with my travis credentials where this builds successfully.

### DIFF
--- a/list/List.js
+++ b/list/List.js
@@ -132,16 +132,14 @@ define([
 		// the SPACE key to (de)select an item.
 		multiCharSearchDuration: 0,
 
-		/**
-		 * Indicates whether the list has a WAI-ARIA role of `listbox` or its default WAI-ARIA role of `grid`.
-		 * If this indicator is set to `true`:
-		 * * The WAI-ARIA role of the list is set to `listbox`;
-		 * * The `selectionMode` property cannot take the value `none` anymore. Its default value becomes `single`;
-		 * * The `itemRenderer` and `categoryRenderer` widget are not allowed to provide internal navigation.
-		 * @member {boolean}
-		 * @default false
-		 */
-		isAriaListbox: false,
+		setAttribute: dcl.superCall(function (sup) {
+			return function (attr, value) {
+				sup.apply(this, arguments);
+				if (attr === "role") {
+					this._applyRole(value);
+				}
+			};
+		}),
 
 		/**
 		 * Defines the scroll direction: `"vertical"` for a scrollable List, `"none"` for a non scrollable List.
@@ -161,14 +159,14 @@ define([
 		},
 
 		/**
-		 * Defines the selection mode: `"none"` (not allowed if `isAriaListbox` is true), `"radio"`, `"single"`
+		 * Defines the selection mode: `"none"` (not allowed if `role=listbox`), `"radio"`, `"single"`
 		 *  or `"multiple"`.
 		 * @member {string} module:deliteful/list/List#selectionMode
-		 * @default "none", or "single" if isAriaListbox is true.
+		 * @default "none", or "single" if `role=listbox`.
 		 */
 		_setSelectionModeAttr: dcl.superCall(function (sup) {
 			return function (value) {
-				if (this.isAriaListbox && value === "none") {
+				if (this.getAttribute("role") === "listbox" && value === "none") {
 					throw new TypeError("selectionMode 'none' is invalid for an aria listbox, "
 							+ "keeping the previous value of '" + this.selectionMode + "'");
 				} else {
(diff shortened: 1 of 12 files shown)


list/List.js
```diff
@@ -132,16 +132,14 @@ define([
...
-		 * * The `itemRenderer` and `categoryRenderer` widget are not allowed to provide internal navigation.
-		 * @member {boolean}
-		 * @default false
-		 */
-		isAriaListbox: false,
+		setAttribute: dcl.superCall(function (sup) {
+			return function (attr, value) {
+				sup.apply(this, arguments);
+				if (attr === "role") {
+					this._propagateRole(value);
```
cjolif: a more common naming than "propagate" might be "_applyRole"  (shorter as well).

list/List.js
```diff
@@ -227,9 +225,14 @@ define([
 		 * @private
 		 */
 
-		render: function () {
+		postRender: function () {
```
cjolif: can you remind me why you had to change this from render to postRender?

list/List.js
```diff
@@ -467,13 +467,14 @@ define([
...
 		/*jshint maxcomplexity:12*/
-		_refreshAriaListboxProperty: function () {
-			this.setAttribute("role", this.isAriaListbox ? "listbox" : "grid");
-			if (this.isAriaListbox) {
+		_propagateRole: function (role) {
+			if (role === "listbox") {
 				if (this.selectionMode === "none") {
 					this.selectionMode = "single";
 				}
 				// TODO: SHOULD WE REMOVE THE FOLLOWING CODE FOR OPTIMIZATION ?
```
cjolif: I think we said we finally don't want to remove this? As switching is common so maybe we should remove that comment?

cjolif:
@tkrugg any feedback on my comments?

list/List.js
tkrugg: ok, I'll make that change.

list/List.js
tkrugg: ok, i'll remove the todo

list/List.js
tkrugg: inside _propagateRole, there is a `this.selectionMode = "single";`, and then inside refreshRendering there is a bunch of things that need to be triggered once this prop has changed. 
In order for selectionMode to be part of the next refresh cycle it needs to be outside render.
@wkeese am I doing something wrong?

wkeese: I think you are right, because delite/Widget doesn't call `initializeInvalidating()` until `postRender()`.

wkeese:
FYI, I looked over this PR briefly and it seems good.

tkrugg:
I did the changes asked.
It's not building successfully (see [log](https://travis-ci.org/tkrugg/deliteful/builds/47537215)) but I suspect it's not related to this PR, as I *tend* to get the same tests failing with the master branch.

tkrugg:
after offline discussion with @cjolif, there is still work to be done here: I haven't completely unlinked the selectionMode and role states: as you can see it by [theses few lines](https://github.com/tkrugg/deliteful/blob/List/getting-rid-of-isAriaListbox%23210/list/List.js#L476-L478). I'm going to fix it.

cjolif:
> I tend to get the same tests failing with the master branch.

Yes the list-prog-5 test is failing without those. I'm working on fixing the test. There should be a commit pretty soon (the problem seems to be in the tests).

tkrugg:
This is ready. It's still [failing on list-prog-5.html](https://travis-ci.org/tkrugg/deliteful/builds/47638428) but as you can see this PR doesn't bring any change to it so there is a chance it won't conflict with the fix @cjolif is working on.

cjolif:
merged in 1924b1e7696afcb4da4ffdd1a55fea14c59bd6f2

Issue #366, opened 2014-10-23
deliteful/ComboBox: revisit the multi-channel policy


AdrianVasiliu:
The current implementation of deliteful/ComboBox displays the popup in the following multi-channel responsive manner:

* "desktop-like": 
  * popup below/above the root node
  * when autoFilter=true (for selectionMode="single"), the `<input>` inside the root node serves for entering filter criteria
  * in selectionMode="multiple", the popup doesn't close as soon as you select an item (such that you can select several), thus the only way to close it is to click outside the popup (or keyboard shortcut in the future).
    * For comparison, browsers show the native HTML `<select multiple>` always in expanded state.
* "phone-like"
  * popup centered (maximizes the availalble area to make the scrolling experience smoother)
  * when autoFilter=true (for selectionMode="single"), an `<input>` is added to the centered popup, such that the user can see it while allowing the popup to obscure the dropdown button.
  * when selectionMode="multiple", OK and Cancel buttons are added to the popup.

The points to be revisited:
* Choice criteria
   * Currently: `has("ios") || has("android"). The user can override a method to define its own choice criteria.
   * Should we treat (large) tablets as desktop? Should viewport size matter?
* How to expose / allow the user to configure the choice criteria between desktop-like and phone-like
    * Currently, on ComboBox, a protected method that can be overridden (`useCenteredDropDown()`).
    * Future: see @cjolif's https://github.com/ibm-js/deliteful/pull/344#discussion_r19021244
* Ergonomy of the popup in "desktop-like" mode
  * Would it be better to add OK - Cancel buttons to the popup in multi-choice mode? (Currently, the user needs to click outside the popup to close the popup - differently than in single-choice mode where the popup closes as soon as you click an option.)
  * Does it make sense to (optionally?) go as browsers go with the native HTML `<select multiple>`, that is show it directly in expanded state?
    * Note that this is how it goes for deliteful/Select (which can't go differently since it leverages the native `<select>`).

Issue #179, opened 2014-07-10
Search/FilterBox element


cjolif:
This should include integration with List as FilteredList

AdrianVasiliu:
Some comments at https://github.com/ibm-js/deliteful/pull/449 might be relevant for the design of this new widget(s) and for their integration with Combobox.

AdrianVasiliu:
PS I refer in particular to these comments from #449:
* https://github.com/ibm-js/deliteful/pull/449#issuecomment-69435625
* https://github.com/ibm-js/deliteful/pull/449#issuecomment-69531579
* https://github.com/ibm-js/deliteful/pull/449#issuecomment-69568366

wkeese:
Closing this since we have no current plans to implement it.

PR #449, opened 2015-01-09
Package centered popup as distinct widget. Fixes #367

### LINKED ISSUE #367: ComboBox: package the popup as a distinct widget
Packaging as a distinct widget the popup created by ComboBox would have at least the advantage that it would make it templatable, thus easier to configure.

When these new widgets will be ready, it should leverage TextBox (#175) or, better, SearchBox/FilterBox (#179). For the time being, it can keep using the native HTML `<input>`.

Notice that ComboBox currently creates two distinct popups, per its multi-channel policy. The distinct widget would need to be able to do both (optionally).

AdrianVasiliu:
Implementation of #367.

To avoid mixing in this PR unrelated changes, I will do further changes separately:
* Cope with the recent changes in delite/FormValueWidget (https://github.com/ibm-js/delite/commit/00f0b74736e4b1a9473ef0c1a1d8ed5971f15bcd ). Since this change, Combobox lost its ability to handle the readOnly attribute of the input element, because now FormValueWidget's own handling of this attribute has the last word on it.
* Styling improvements to improve the rendering (including for vertical alignment of text and arrow in IE, and rendering of inputs of type="search").
* And of course Combobox' API for creating the popup will change again with the implementation of #366 (multi-channel policy) and the overall design of Combobox/ComboPopup will be impacted by #179 (Seach/FilterBox / FilteredList).

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -9,13 +9,12 @@ define([
 	"delite/HasDropDown",
 	"delite/keys",
 	"./list/List",
-	"./LinearLayout",
-	"./Button",
+	"./Combobox/ComboPopup",
 	"delite/handlebars!./Combobox/Combobox.html",
 	"requirejs-dplugins/i18n!./Combobox/nls/Combobox",
 	"delite/theme!./Combobox/themes/{{theme}}/Combobox.css"
 ], function (dcl, $, Filter, has, register, FormValueWidget, HasDropDown,
-		keys, List, LinearLayout, Button, template, messages) {
+		keys, List, ComboPopup, template, messages) {
 	/**
 	 * A form-aware and store-aware widget leveraging the `deliteful/list/List`
 	 * widget for rendering the options.
@@ -198,11 +197,23 @@ define([
 		 */
 		multipleChoiceNoSelectionMsg: messages["multiple-choice-no-selection"],
 		
-		// TODO: worth exposing public properties?
-		// The default label of the OK button
-		_okButtonLabel: messages["ok-button-label"],
-		// The default label of the Cancel button
-		_cancelButtonLabel: messages["cancel-button-label"],
+		/**
+		 * The text displayed in the OK button when the combobox popup contains such a button.
+		 * The default value is provided by the "ok-button-label" key of
+		 * the message bundle.
+		 * @member {string}
+		 * @default "OK"
+		 */
+		okMsg: messages["ok-button-label"],
+		
+		/**
+		 * The text displayed in the Cancel button when the combobox popup contains such a button.
+		 * The default value is provided by the "cancel-button-label" key of
+		 * the message bundle.
+		 * @member {string}
+		 * @default "Cancel"
+		 */
+		cancelMsg: messages["cancel-button-label"],
 		
 		preRender: function () {
 			this.list = new List();
@@ -330,7 +341,7 @@ define([
 			this.list.selectionMode = this.selectionMode === "single" ?
 				"radio" : "multiple";
 			
-			var dropDown = this._createDropDown(this.list);
+			var dropDown = this._createDropDown();
 			
 			// Since the dropdown is not a child of the Combobox, it will not inherit
 			// its dir attribute. Hence:
@@ -467,15 +478,15 @@ define([
 			return has("ios") || has("android");
 		},
 		
-		_createDropDown: function (list) {
+		_createDropDown: function () {
 			// Update the readonly attribute in case useCenteredDropDown() changed
 			// its return value.
 			this._updateInputReadOnly();
 			
 			var centeredDropDown = this.useCenteredDropDown();
 			var dropDown = centeredDropDown ?
-				this._createCenteredDropDown(list) :
-				this._createNormalDropDown(list);
+				this.createCenteredDropDown() :
+				this.createAboveBelowDropDown();
 			
 			this.dropDownPosition = centeredDropDown ?
 				["center"] :
@@ -487,69 +498,31 @@ define([
 			return dropDown;
 		},
 		
-		_createNormalDropDown: function (list) {
+		/**
+		 * Factory method which creates the widget used inside above/below drop-down.
+		 * The default implementation simply returns `this.list`.
+		 * @protected
+		 */
+		createAboveBelowDropDown: function () {
 			// Use the List itself as content of the popup. Embedding it in a
 			// LinearLayout has seemed useful for solving layout issues on iOS
 			// (deliteful issue #270), but appears to be harmful on IE11 (deliteful
 			// issue #382). Hence the List is not wrapped anymore inside a LinearLayout.
-			return list;
-		},
-		
-		_createCenteredDropDown: function (list) {
-			// TODO: move to separate widget.
-			var topLayout = new LinearLayout();
-
-			if (this.autoFilter && this.selectionMode !== "multiple") {
-				this._popupInput = this._createPopupInput();
-				topLayout.addChild(this._popupInput);
-			}
-			
-			$(list).addClass("fill");
-			topLayout.addChild(list);
-
-			// Just as Android for the native select element, only use ok/cancel
-			// buttons in the multichoice case.
-			if (this.selectionMode === "multiple") {
-				var bottomLayout = new LinearLayout({vertical: false, width: "100%"});
-				var cancelButton = new Button({label: this._cancelButtonLabel});
-				var okButton = new Button({label: this._okButtonLabel});
-				okButton.onclick = function () {
-					this._validateMultiple(this.inputNode);
-					this.closeDropDown();
-				}.bind(this);
-				cancelButton.onclick = function () {
-					this.list.selectedItems = this._selectedItems;
-					this.closeDropDown();
-				}.bind(this);
-				$(cancelButton).addClass("fill");
-				$(okButton).addClass("fill");
-				bottomLayout.addChild(cancelButton);
-				bottomLayout.addChild(okButton);
-				topLayout.addChild(bottomLayout);
-			}
-			return topLayout;
+			return this.list;
 		},
 		
 		/**
-		 * Creates the input element inside the popup.
-		 * Only used for single-choice mode.
-		 * @private
+		 * Factory method which creates the widget used inside centered drop-down.
+		 * The default implementation returns a new instance of deliteful/Combobox/ComboPopup
+		 * (the present widget is set for its `combobox` property).
+		 * The method can be overridden in order to create a subclass of ComboPopup (for
+		 * specifying a custom template, for instance).
+		 * @protected
 		 */
-		_createPopupInput: function () {
-			// TODO: use deliteful/SearchBox when will be available.
-			var popupInput = document.createElement("input");
-			$(popupInput).addClass("d-combobox-popup-input");
-			popupInput.setAttribute("role", "combobox");
-			popupInput.setAttribute("autocomplete", "off");
-			popupInput.setAttribute("autocapitalize", "none");
-			popupInput.setAttribute("autocorrect", "off");
-			popupInput.setAttribute("aria-autocomplete", "list");
-			popupInput.setAttribute("type", "search");
-			popupInput.setAttribute("placeholder", this.searchPlaceHolder);
-			this._prepareInput(popupInput);
-			return popupInput;
+		createCenteredDropDown: function () {
+			return new ComboPopup({combobox: this});
 		},
-
+		
 		_prepareInput: function (inputElement) {
 			this.on("input", function (evt) {
 				// Would be nice to also have an "incrementalFilter" boolean property.
--- a/Combobox/ComboPopup.html
+++ b/Combobox/ComboPopup.html
@@ -1,0 +1,19 @@
+<template role="presentation"
+	requires="deliteful/LinearLayout, deliteful/Button">
+	<d-linear-layout style="height: 100%">
+		<input d-hidden="{{!(this.combobox.autoFilter && this.combobox.selectionMode !== 'multiple')}}"
+			attach-point="inputNode"
+			class="d-combobox-input"
+			role="combobox" autocomplete="off" autocapitalize="none" autocorrect="off"
+			aria-autocomplete="list" type="search"
+			placeholder="{{combobox.searchPlaceHolder}}">
+		<div attach-point="listNode"></div>
+		<d-linear-layout d-hidden="{{combobox.selectionMode !== 'multiple'}}"
+			vertical="false">
+			<button is="d-button" class="fill d-combo-cancel-button"
+				label="{{combobox.cancelMsg}}" on-click="{{cancelHandler}}"></button>
+			<button is="d-button" class="fill d-combo-ok-button"
+				label="{{combobox.okMsg}}" on-click="{{okHandler}}"></button>
+		</d-linear-layout>
+	</d-linear-layout>
+</template>
--- a/Combobox/ComboPopup.js
+++ b/Combobox/ComboPopup.js
@@ -1,0 +1,69 @@
+/** @module deliteful/ComboPopup */
+define([
+	"delite/register",
+	"requirejs-dplugins/jquery!attributes/classes",	// addClass()
+	"delite/Widget",
+	"delite/handlebars!./ComboPopup.html"
+], function (register, $, Widget, template) {
+	/**
+	 * Auxiliary widget used in some cases by deliteful/Combobox for displaying
+	 * a popup containing conditionally a search field and OK/Cancel buttons.
+	 * This widget is intended for being instanciated only by deliteful/Combobox;
+	 * it should not be instanciated directly. If needed, its template
+	 * (deliteful/Combobox/ComboPopup.html) can be customized.
+	 * @class module:deliteful/Combobox/ComboPopup
+	 * @augments module:delite/Widget
+	 */
+	return register("d-combo-popup", [HTMLElement, Widget],
+		/** @lends module:deliteful/ComboPopup# */ {
+		
+		baseClass: "d-combo-popup",
+		
+		template: template,
+		
+		/**
+		 * The instance of `deliteful/Combobox` for which this widget is used.
+		 * This property is set by Combobox when creating the popup, and it
+		 * is used in the template of ComboPopup for accessing the properties
+		 * of the Combobox.
+		 * @member {boolean} module:deliteful/Combobox/ComboPopup#combobox
+		 * @default null
+		 * @protected
+		 */
+		combobox: null,
+		
+		refreshRendering: function (oldValues) {
+			if ("combobox" in oldValues) {
+				if (this.combobox) {
+					var list = this.combobox.list;
+					if (list) {
+						list.placeAt(this.listNode, "replace");
+						// startup() is not called by Widget.placeAt because the parent of
+						// dropDown is not yet started, but startup() will be called by delite/popup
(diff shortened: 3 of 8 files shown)


Combobox.js
```diff
@@ -9,13 +9,12 @@ define([
 	"delite/HasDropDown",
 	"delite/keys",
 	"./list/List",
-	"./LinearLayout",
-	"./Button",
+	"./Combobox/ComboPopup",
```
cjolif: As we are already in the ComboBox directory, and for the sake of not duplicating naming information already conveyed, I'm not sure if we have to restate "Combo" before Popup.

AdrianVasiliu: Maybe. On the other side, it somehow makes the scope of this module clearer, including when looking just at the file name. And we already have redundantly Combobox/Combobox.html (the template, which could be called generaically template.html). 

All in one:
* I'm willing to change the name, if there's agreement this would be better.
* Personally, I think both could work, while having a small preference on keeping ComboPopup rather than just Popup.
* Since you seem to hesitate on it too, I'm not sure what to do. For now, I wait for further feedback.

wkeese:
I think if this widget is designed well, it can be used independently of ComboBox, and also not necessarily as a popup (just like dijit/Calendar).   So I say to:

* Instead of having a combobox property, have properties placeHolder, okMsg, cancelMsg, showButtons, showInput ...
* Name it something like FilteredSelect.   Not sure what name is best, but I'm hoping for a name that reflects what the widget itself does, rather than how it's used by ComboBox.

AdrianVasiliu:
@wkeese Before going with the current design, I first thought at something around what you describe. Here are some reasons for which I finally did not go with a generic widget:

* The input element and the 2 buttons (OK and Cancel) do not serve any purpose as long as there is no action defined for them. So, to be used for something else than for Combobox, not only the properties placeHolder, okMsg, cancelMsg, and the boolean flags showButtons/showInput would need to be set, but also the event handlers, for which a generic default implementation does not make sense. Without predefined actions, a generic widget which just puts together a couple of elements seems quite pointless and overkill to me.
* The structure of ComboPopup fits the usage in ComboBox. I don't see much reason for a generic widget to have the input necessarily on top, unless it's about an input used for filtering the List placed below the input as in the typical case of a filtered list. To be really generic, I guess it would need to allow placing the input anywhere, by means of a position property, not just by the heavy mean of customizing the template, which requires subclassing.
* The combination of a List and an input field used for filtering the List does make sense generically, and is already covered by #179 (say, a FilteredList widget). What we (@cjolif and myself) had in mind was that, if/when we'll have this widget, ComboPopup would leverage it (while adding the OK/Cancel buttons). That is, the existence of a non-generic ComboPopup is not exclusive with the introduction of other widgets serving more generic purposes. ComboPopup would reuse and specialize it for the needs of ComboBox.
* As I told you in our chat some time ago, ComboPopup was initially intended to be non-documented, precisely because it is not more than an auxiliary widget for Combobox. The initial motivation for packaging it distinctly from Combobox was to allow it to be templated (thus, allowing users to customize the template)  - the templating mechanism not allowing a widget to have distinct templates for different parts of its GUI. Then, ComboPopup became documented because it doesn't make sense that users customize the template of an undocumented widget...

I do not pretend these reasons close the debate. I just mention them to feed the brainstorming.

wkeese:
I'm only suggesting simple changes, not complex things like having a property to reposition the `<input>` node. 

My initial suggestion is very simple: make all the properties scalar, like they should be anyway because that's how other Elements (see the HTML spec) and other custom elements (look at deliteful) work whenever possible.  So Combobox would create the popup like:

```js
var popup = new ComboPopup({
	placeHolder: this.searchPlaceholder,
	list: this.list,
	okMsg: this.okMsg,
	cancelMsg: this.cancelMsg,
	showInput: this.autoFilter && this.selectionMode !== 'multiple',
	showButtons: this.selectionMode === 'multiple'
});
```

It's true I forgot about the event handlers for OK/Cancel buttons in my original comment, but it's simple and cleaner to also handle those event from Combobox, rather than inside ComboPopup.  So ComboPopup would just fire the events, and Combobox would do:

```js
popup.on("ok", function () {
	this._validateMultiple(this.inputNode);
	this.closeDropDown();
}.bind(this));
popup.on("cancel", function () {
	this.list.selectedItems = this._selectedItems;
	this.combobox.closeDropDown();
}.bind(this));
```

Other notes:

* I just looked at `_createPopupInput()` -- Shouldn't that be part of this ComboPopup rather than part of Combobox?
* Also, about `_prepareInput()`, is that being used for both when there is and when there isn't a popup?

AdrianVasiliu:
> I'm only suggesting simple changes, not complex things like having a property to reposition the <input> node. My initial suggestion is very simple: make all the properties scalar

I think the API design needs to be done after, not before deciding if this widget should be a specialized auxiliary widget for Combobox, or a more generic one. If the widget would become generic, then for sure it would have scalar properties and would fire events, as you say, instead of having predefined event handlers and having a "combobox" property for getting access to the properties and methods of Combobox used in the template and in the code of ComboPopup. But yes, if we decide to convert it into a generic widget, this is how the API would be. 

Now, as already said, in my eyes a more generic widget is overkill and has little plus-value for users. If the widget becomes generic and is no longer documented as an auxiliary of Combobox, then it would need to be moved from inside the Combobox directory to the deliteful top-level, and have its own theme (instead of sharing its theme with Combobox), doc, samples etc. I can hardly see how its tiny features and value can justify that. And, if the widget remains specific to Combobox, in my eyes the current API fits better its current specific purpose/scope, because it avoids the useless duplication of scalar properties between Combobox and ComboPopup and the need for Combobox to keep in sync its own scalar properties with those of ComboPopup. If we decide to change the purpose/scope of the widget, then yes, the API would change, no doubt... 

So, I don't think these changes are tricky at all, but we only need them if we want a more generic widget, and if we really feel this would bring plus-value.

> It's true I forgot about the event handlers for OK/Cancel buttons 

(The same holds for the handler of "input" events of the input element. There's no meaningful predefined action that could be executed by a generic widget when the user types into the input element).

> I just looked at `_createPopupInput()` -- Shouldn't that be part of this ComboPopup rather than part of Combobox?

There's no `_createPopupInput()` in Combobox.js from my PR (https://github.com/AdrianVasiliu/deliteful/blob/combo-popup-widget-pr/Combobox.js). I guess you looked in the older branch (now obsolete) of my fork instead of the branch (combo-popup-widget-pr) used by the PR.

> Also, about `_prepareInput()`, is that being used for both when there is and when there isn't a popup?

Yes. And this is why the method is in Combobox, not in ComboPopup.

wkeese:
> I think the API design needs to be done after, not before deciding if this widget should be a specialized auxiliary widget for Combobox, or a more generic one.

According to you, you've already decided to first make a specialized widget and then later replace it with a general widget.   That seems a bit silly to me, but since you've already decided, let's go with that.

Even though you are (initially) making a specialized helper widget, the changes I suggested still make sense, because they are based on the general programming principle of keeping classes (or in this case, widgets) as independent as possible, to avoid spaghetti code.   That's especially important in this case since you've said that the whole purpose of the ComboPopup widget is for users to customize it.  If users are going to be working directly with it, the API should be as clean as possible.

AdrianVasiliu:
> According to you, you've already decided to first make a specialized widget and then later replace it with a general widget.

I didn't say that exactly. I said two things:
* We decided (wrongly or not, this is debatable) to use a specialized widget for the popup of Combobox.
* And we consider adding later a more generic widget for List + search field. Once this will exist, we'll refactor ComboPopup to leverage it while specializing it by the addition of OK/Cancel buttons. This is different than *replacing* ComboPopup with the future SearchList.

> Even though you are (initially) making a specialized helper widget, the changes I suggested still make sense, because they are based on the general programming principle of keeping classes (or in this case, widgets) as independent as possible, to avoid spaghetti code.

I'm not sure that accessing in ComboPopup the properties of Combobox via a `combobox` property is spaghetti code. Again, this design wouldn't be good if ComboPopup would be a first-level widget, usable standalone, but as long as it isn't, the current API has some advantages. Why wouldn't we think it's a pitty to duplicate a bunch of properties on both Combobox and ComboPopup and to add code to keep them in sync, as long as the later is just a helper widget for the first widget? 

> That's especially important in this case since you've said that the whole purpose of the ComboPopup widget is for users to customize it. If users are going to be working directly with it, the API should be as clean as possible.

I do agree the current design is a bit suboptimal, but I think the deep reason is that the current templating mechanism forces us to create a distinct widget just to be able to use a template for part of the widget which is not inside the root node. And this suboptimality is to be compared with the suboptimality that would would exist if we'd duplicate properties. Also, when I refered to users customizing ComboPopup, I mostly refered to customizing the template. The use-cases that I considered realistic consist in users adding, say, a title or somehow altering the structure of the popup, at the level of the template, not that much at the level of the API of ComboPopup.

wkeese:
Well, ComboPopup is definitely spaghetti code.  If it were clean, you could use and test ComboPopupt independently of Combobox.

AdrianVasiliu:
I'm open to make changes and improve it. But for considering it I'd first need to be sure I understand correctly your vision.

You initially said the widget should not only be designed to be technically usable without Combobox, but also to be named (and documented, I assume) in a way that "reflects what the widget itself does, rather than how it's used by Combobox" (while you were unsure yourself about what a good name could be). I've answered that I don't see significant use-cases for a usage outside Combobox where it would bring value and therefore promoting this widget from the status of a helper to a first-level deliteful citizen seems unjustified to me. I'm still unclear if you do see such use-cases, and when you speak about making it independent of Combobox I'm unclear if by that you still imply it should be packaged as a first-class deliteful widget, thus adding for it independent theme, doc, samples, and tests, while it currently shares them with Combobox. 

That is, before considering a rewrite, I would need to understand how to justify its existence as an independent widget, and what would be its requirements / feature set / use-cases.

wkeese:
>  I'm still unclear if you do see such use-cases, and when you speak about making it independent of Combobox

I'm not sure.   Anyway, I'm not interested in debating that point.

>  I'm unclear if by that you still imply it should be packaged as a first-class deliteful widget, thus adding for it independent theme, doc, samples, and tests, while it currently shares them with Combobox.

Not sure, it seems like it would make sense for the CSS to be independent, but it might be overkill to have separate tests and samples.

> I would need to understand how to justify its existence as an independent widget, and what would be its requirements / feature set / use-cases.

As I've said before, making classes independent is a basic programming principle, regardless of whether or not they will be used independently.  Some of the reasons are:

* code reuse
* ease of testing
* ease of maintenance
* ease of understanding the code
* consistency with other code
* ease of extending the code (subclassing)

These principles apply even for private classes.

In addition, following the standard rules and conventions for writing widgets -- how widget properties are declared, and how widgets emit events -- also leads to ease of testing,  and ease of understanding the code, etc.

You don't seem to think that any of these reasons are compelling.  If so, there's no point in continuing the discussion.

AdrianVasiliu:
>> I'm still unclear if you do see such use-cases [independently of Combobox], and when you speak about making it independent of Combobox

> I'm not sure. Anyway, I'm not interested in debating that point.

You raised this point initially, isn't it? You also refered to a renaming to reflect a more generic usage. I took it seriously, that's why I was trying to clarify it and hopefully reach a consensus, before taking action if appropriate.

> You don't seem to think that any of these reasons are compelling. If so, there's no point in continuing the discussion.

I thought that by writing "I'm open to make changes and improve it. But for considering it I'd first need to be sure I understand correctly your vision." I made it clear enough that I'm not closed to your arguments. Anyway, let's give it another look and another chance tomorrow, after a good sleep night on my side.

wkeese:
> You raised this point initially, isn't it?

Yes, but I was just listing one of the many reasons for making ComboPopup independent.   As I've said repeatedly, make ComboPopup independent is good programming even if it will never be used standalone.

To be honest, I was just listing the trivial changes to get ComboPopup to be more independent.   I think for it to be designed correctly, the desktop version of Combobox should share a common base class with ComboPopup.   After all, they basically do the same thing (which is why they both have an `<input>` that filters a List).

AdrianVasiliu:
>  I think for it to be designed correctly, the desktop version of Combobox should share a common base class with ComboPopup.   After all, they basically do the same thing (which is why they both have an `<input>` that filters a List).

* Providing a class shared between Combobox and ComboPopup is part of the current plan. As already mentioned, the current plan for the next milestone is to introduce a FilteredList widget - #179 (or FilterableList, or another name that will be decided). This contains an `<input>` that filters a List. Its API will likely include at least `searchPlaceHolder` and `list` properties, and the code and API for executing the filtering will move from Combobox to this new widget. That's the rough idea as I have it in mind, but it may evolve when we'll start working on it. I agree there would have been advantages to proceed in the reversed order, or, even better, to design everything altogether. But we did Combobox first because it looked higher priority, while the FilteredList was mostly an oportunity to reuse/repackage code written already for Combobox to get for a low cost a widget providing the feature outside of Combobox context. 
* You mention that the desktop version of Combobox would share a common base class with ComboPopup. Only the desktop version? Would we have a different inheritance depending on the selected multi-channel policy? In my mind, the list filtering capability is exactly the same regardless of the channel policy (say, desktop as for mobile). The `<input>` used for filtering behaves the same (maybe with some styling differences). What changes is the DOM structure. For desktop, the FilteredList goes directly under the root node of Combobox, while for mobile it gets embedded in the ComboPopup widget. Making Combobox inherit (regardless of channel policy) from FilteredList would be a way to avoid duplicating API, but this would need to be examined more closely when work will start on it.

wkeese:
> You mention that the desktop version of Combobox would share a common base class with ComboPopup. Only the desktop version? Would we have a different inheritance depending on the selected multi-channel policy? 

Well, it seems a bit silly for the mobile Combobox to pull in all that code when all it does it to basically [load and] display the ComboPopup.   It's essentially a ToggleButton.  OTOH sharing a base class would get all the public fields (like store, selectionMode, etc.), so I don't know.

AdrianVasiliu:
> OTOH sharing a base class would get all the public fields (like store, selectionMode, etc.), so I don't know.

The properties `store` and `selectionMode` exists on Combobox and List, but not on ComboPopup (and they wouldn't be there even if it gets refactored as you suggested), so I don't know how a common base class between Combobox and ComboPopup would change anything with this respect. Maybe I misunderstand your point?

AdrianVasiliu:
Also, another aspect in case it might matter: as you know, the implementation of the multi-channel policy (#366) as it is currently considered relies on Combobox conditionally requiring the ComboPopup widget. Something around:

Combobox.js:
```js
define(["...", "requirejs-dplugins/has!centered-combo-popup?ComboPopupCentered:ComboPopupAboveBelow"], function(..., ComboPopup) {
   register("element", [..., ComboPopup], {
   ... 
   });
}
```

That is, in the desktop policy, Combobox would not require ComboPopup.

AdrianVasiliu:
Let me try to summarize my understanding of where we are.
* ComboPopup is currently designed as helper for Combobox and currently has the `combobox`(non-scalar) property. 
* Concerning your initial comment that you'd like to see scalar properties on ComboPopup instead of the `combobox` property and events fired for OK/Cancel buttons instead of predefined event handlers: I have mixed feelings about it. While I do understand that designing a ComboPopup such that it can be used independently on Combobox has advantages even if Combobox remains its only use-case, I also see some reasons to stay for now with it as it is:
  * Since as already said there will be anyway a refactoring when FilteredList will exist and the details of this refactoring are not yet worked out, the current very short API of ComboPopup seems easier to refactor / less likely to lead to incompatible changes (besides avoiding the duplication of the API between Combobox and ComboPopup). In principle, we might also see things clearer once we're done with #366 (multi-channel policy).
  * (for what this argument is worth) The current ComboPopup is pretty much similar (with this respect) to the situation and design of https://github.com/ibm-js/deliteful/blob/master/ViewIndicator.js (which has a `viewstack` property).

All in one, I think it wouldn't be completely non-sense if we'd go at this stage with ComboPopup as it is in the PR, while doing a global rethinking of the design and API at the occasion of #179 (FilteredList), that is one all the pieces of the puzzle will be in place.

Would that be acceptable for you?

wkeese:
I'm tired of arguing about it, so anything is acceptable to me.

About the inheritance changes, they would become obvious to you if you tried to make ComboPopup an independent widget (without duplicating code between ComboPopup and Combobox).

AdrianVasiliu:
> I'm tired of arguing about it, so anything is acceptable to me.

I would be happy if we could reach a shared vision without getting anybody tired, and I'm sorry that this discussion got you tired. But I don't see a better way, other than exchanging views about technical matters. Personally, I do keep listening your arguments, and trying to better understand / clarify them.

> About the inheritance changes, they would become obvious to you if you tried to make ComboPopup an independent widget (without duplicating code between ComboPopup and Combobox).

Did I say there will be no inheritance changes? I said the picture will change anyway at the next step, when a FilteredList widget will exist (how exactly being a matter of thinking once this future widget will exist). Here, I was answering your point:

> OTOH sharing a base class would get all the public fields (like store, selectionMode, etc.)

and for `store` and `selectionMode` I mentioned that they can't be shared via a common base class between Combobox and ComboPopup because `store` is not present on any of them and `selectionMode` is only present on one of them. 

Yes, other properties (okMsg etc.) and methods are likely to be shared via a common base class once FilteredList will exist. But I did feel and still feel overkill to introduce now (before FilteredList gets into the picture) yet another (base) class.

cjolif:
Considering the limited time we can spend on deliteful overall vs the amount of theoretical work involved I think we must adopt an agile and pragmatic approach. That means build reasonable steps for each milestone and improve them at next milestone. Current status is not perfect but as @AdrianVasiliu is saying he is planning for improving it later I think it makes sense to wait for that moment so that we have "all the cards in hands" to do that series of improvements. Maybe @AdrianVasiliu you can merge this but save the comments from @wkeese in the ticket corresponding to the additional work so that there are not forgotten.

AdrianVasiliu:
Okay, so I pushed it via 8bb3febccfc4664f017659918958ccb9e93f76f8, and pointed in #179 to @wkeese comments which seemed relevant.

AdrianVasiliu:
@wkeese, just noticed that you added a PS to a [previous comment](https://github.com/ibm-js/deliteful/pull/449#issuecomment-69568366), which holds for a design point unrelated with the present PR but it's worth discussing it too, so let me answer it:

> PS: One thought I had was to speed up page load time on mobile by deferring loading the List widget etc. until someone clicks the Combobox. You could follow the same strategy on desktop, but it works even better on mobile because everything is inside the dialog.

I did thought about (optional) lazy loading, but here are some points that seem to make it inapplicable or hardly applicable for Combobox:

* As a general thought, the usefulness of lazy loading of the List would vary depending on applications. An app with many combos that are unlikely to be ever open would of course gain, but other apps would consider initial loading time less critical for the global perceived responsiveness of the app than how quick the popup opens once the user clicks the combo. This holds including for mobile, where I think people are used with a little bit of wait at app launch but can't accept a waiting time after touching a button.
* More particularly, the List itself offers already means to get some kind of lazy loading, via PageableList. I'm not sure how much it makes sense to use a PageableList with Combobox, nor how well PageableList currently behaves (it could be matter of future work to dig into it), but this is one possible path to follow.
* Finally, both Combobox and Select, consistently with the native `<select>`, show the label of the first option even before the user clicks the button, which implies that at least the first List item is already loaded before the click. Okay, we can imagine that this label is provided by the user instead of getting it from the List in the case of lazy loading, but this would (partially) defeat the principle that data comes from the store. Still, we could consider it, if demand will exist, I would say.

Issue #450, opened 2015-01-12
Switch: support textDir property


semion1956:
Support textDir for Switch labels (checkedLabel, uncheckedLabel) as well as title.

cjolif:
see aso is #454 for other possible rtl issues with switch.

semion1956:
This issue is about controlling text direction in the labels of Switch. #454 is about css, which doesn't cause rtl problem, but probably can be written better. As far as I can see, it isn't connected with current issue and should be solved separately.

cjolif:
@semion1956 that's perfectly fine solving them separately just wanted to make to link both together in case one influence the other.

PR #451, opened 2015-01-12
Bidi problems in Switch

semion1956:
Changes to fix issue #450

### DIFF
--- a/Switch/bidi/Switch.js
+++ b/Switch/bidi/Switch.js
@@ -5,11 +5,14 @@ define([
 
 	return dcl(null, {
 		refreshRendering: function (oldVals) {
-			if ("textDir" in oldVals || "checkedLabel" in oldVals) {
+			if ("dir" in oldVals || "textDir" in oldVals || "checkedLabel" in oldVals) {
 				this._innerNode.firstChild.textContent = this.applyTextDirection(this.checkedLabel);
 			}
-			if ("textDir" in oldVals || "uncheckedLabel" in oldVals) {
+			if ("dir" in oldVals || "textDir" in oldVals || "uncheckedLabel" in oldVals) {
 				this._innerNode.lastChild.textContent = this.applyTextDirection(this.uncheckedLabel);
+			}
+			if (this.title && "textDir" in oldVals) {
+				this.title = this.applyTextDirection(this.title);
 			}
 		},
 
--- a/Switch/bidi/Switch.js
+++ b/Switch/bidi/Switch.js
@@ -5,16 +5,11 @@ define([
 
 	return dcl(null, {
 		refreshRendering: function (oldVals) {
-			if (this.checked) {
-				if ("checked" in oldVals || "textDir" in oldVals || "checkedLabel" in oldVals) {
-					this._innerNode.firstChild.textContent = this.applyTextDirection(this.checkedLabel);
-				}
+			if ("textDir" in oldVals || "checkedLabel" in oldVals) {
+				this._innerNode.firstChild.textContent = this.applyTextDirection(this.checkedLabel);
 			}
-			else if ("checked" in oldVals || "textDir" in oldVals || "uncheckedLabel" in oldVals) {
+			if ("textDir" in oldVals || "uncheckedLabel" in oldVals) {
 				this._innerNode.lastChild.textContent = this.applyTextDirection(this.uncheckedLabel);
-			}
-			if (this.title && "textDir" in oldVals) {
-				this.title = this.applyTextDirection(this.title);
 			}
 		},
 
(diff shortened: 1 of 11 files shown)


Switch/bidi/Switch.js
```diff
@@ -1,19 +1,34 @@
...
 ], function (dcl) {
-	// module:
-	//		deliteful/Switch/bidi/Switch
 
 	return dcl(null, {
-		_setCheckedLabelAttr: function (value) {
-			value = this.wrapWithUcc(value);
-			this._set("checkedLabel", value);
+		refreshRendering: function (oldVals) {
+			var div;
```
cjolif: as this is starting to be a little bit more than just a couple of lines of code we might want to do them inside a has("bidi") flag to scope it to use-cases where it is needed?

cjolif: sorry we are in bidi module so what I say does not make sense.

cjolif:
You should move the bidi test to the new structure (tests/unit/bidi) following #433

cjolif:
@semion1956 do you think you'll be able to do the test path update on that one before next week 0.6 deadline?

Not related but also please put space between any word preceding #issuenumber because if you don't put a space the linking is not working as expected.

cjolif:
also please make sure to check jshint (grunt jshint) to avoid problems like: https://github.com/ibm-js/deliteful/commit/e37e77e311f5925833240783353bc32e78ed3a29

tests/unit/Switch-bidi.js
```diff
@@ -0,0 +1,136 @@
...
+	"intern/chai!assert"
+], function (registerSuite, assert) {
+	var container, html = "<d-switch id='b1' checkedLabel='\u05d0\u05d1\u05d2' uncheckedLabel='ABC' checked='true' textDir='ltr'></d-switch>" +
+		"<d-switch id='b2' checkedLabel='\u05d0\u05d1\u05d2' uncheckedLabel='ABC' checked='false' textDir='rtl'></d-switch>" +
+		"<d-switch id='b3' checkedLabel='\u05d0\u05d1\u05d2' uncheckedLabel='ABC' checked='true' title='ABC \u05d0\u05d1\u05d2' textDir='auto'></d-switch>";
+
+	registerSuite ({
+		name: "deliteful/Switch (bidi)",
+		setup: function () {
+			moduleRequire = require.config({
```
cjolif: note that this solution is actually not working, I had to comment out bidi  tests from the repo (see: c32a6d83ce00014628f45b6f8b0ff637efbe0c1f). The reason is that by using context require, modules likes Switch happened to be loaded twice. Once with bidi and once without. This sounds good. The problem is that then we end up registering the d-switch  custom element twice and this is something that is forbidden by custom element specification and fire exceptions at least in Chrome.

cjolif: Thinking more about this, I think the only viable solution will be to have bidi tests totally separated with its own intern config (inheriting from the standard one) and another grunt target to launch bidi tests (and maybe yet another one to run both bidi and non bidi).

wkeese: Hmm, I forgot about that.  Your solution sounds fine, although I think another approach would be to have some kind of clear() method on delite/register.

wkeese: Woops, ignore my suggestion because it won't work for platforms that have a native support for document.registerElement().

cjolif:
@semion1956 any update?

semion1956:
This week I was involved in some other projects, which took all my working time. Updates will be done in the start of the next week.

cjolif:
ok thanks, so that will be for next milestone.

semion1956:
Latest changes also take into account #454 and #467.

wkeese:
> Latest changes also take into account #454

If that's really true then why does the template still have this code in it?

```js
{{this.isLeftToRight() ? this.checkedLabel : this.uncheckedLabel}}
```

Did you push your latest changes?

semion1956:
Done

wkeese:
Thanks, the Switch code is looking much better than it did before.    I just have one comment about refreshRendering() that I'll write inline there.

semion1956:
"checked" in oldVals should not be checked, I updated this.
Title in oldVals (at least currently) can't appear.

wkeese:
OK, thanks.

This PR is corrupted now though.   Looks like you tried to do a "git merge" again.  Can you rebase it to remove all the spurious commits?   I'm especially worried because it now shows changes to Button and ToggleButton, and I'm not sure if they are supposed to be there or not.

semion1956:
I did have some strange problems with latest commit, but 'git merge' was not done this time. However, changes in tests for bidi/Button and bidi/ToggleButton together with changes in tests for bidi/Switch are included in this PR correctly. As per #467 I removed loader configurations 'in place'.

wkeese:
I fixed the PR corruption and the jshint errors and pushed this as 0f9cf7e1de6a71d6aa4e0596d44c40651873c422.

PR #452, opened 2015-01-12
Lie for Toaster message

tkrugg:
- wrapped everything inside new Promise()
- added destroy method to timer so that it's supported by this.own()
- removed check on _promise.state

### DIFF
--- a/ToasterMessage.js
+++ b/ToasterMessage.js
@@ -2,11 +2,11 @@
 define(["dcl/dcl",
 	"delite/Widget",
 	"delite/register",
-	"dojo/Deferred",
+	"lie/dist/lie",
 	"requirejs-dplugins/jquery!attributes/classes",
 	"dpointer/events",
 	"delite/handlebars!./Toaster/ToasterMessage.html"
-], function (dcl, Widget, register, Deferred, $, pointer, template) {
+], function (dcl, Widget, register, Promise, $, pointer, template) {
 
 	// TODO: this could be abstracted in a separate class, so that it can be used by other widgets
 	// such as the toggle/switch.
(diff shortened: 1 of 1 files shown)


tkrugg:
@wkeese I set up a flag _fulfilled to do the same thing. I thought it would be easier to explain why we're not using _remaining == 0.
I also updated the jshintrc to support Promise as a global, I don't know if it's the right thing to do. Feel free to leave that commit out if necessary.

wkeese:
OK.   That's better than before, although I'm not sure why you want a `fulfilled` flag in addition to the `timer` property, especially when the `pause()` method still does:

```js
if (timer !== null)
```

Also, I'm unclear about the Promise global.   We never use Promise as a global and if I understand the lie documentation it doesn't create a global when the module is loaded via AMD.   So, I don't see the purpose of changing jshintrc.

cjolif:
yes what is missing is requiring Lie in ToastMessage.js to import the Promise object...

tkrugg:
The pause method checks if there is a timer defined before it clears it.
The destroy method doesn't really care about that. It just doesn't want to reject a promise if it's been fulfilled. I fount it was more readable that the test happens on a variable called "_fulfilled". it's really just an opinion to me.

> Also, I'm unclear about the Promise global
yes, sorry i'll fix that.

wkeese:
OK, makes sense to me.   One last thing:  when you do clearTimeout(this._timer) it's probably a good idea to set this._timer to null.

tkrugg:
it's done

cjolif:
merged in ba968b7f9b6ed5a2f7ca56c565f9954e683744e4

PR #463, opened 2015-01-21
Multichannel infrastructure and its use by Combobox (and ResponsiveColumns). Fixes #366

### LINKED ISSUE #366: deliteful/ComboBox: revisit the multi-channel policy

AdrianVasiliu:
This PR implements #366 and includes:

* Two new multichannel infrastructure modules:
  * https://github.com/AdrianVasiliu/deliteful/blob/channelPolicy-pr/features.js
    * Extends `has()` with channel features (phone-like, tablet-like, and destop-like) computed using CSS media queries based on screen width.
    * Used by multichannel widgets (Combobox for now).
  * https://github.com/AdrianVasiliu/deliteful/blob/channelPolicy-pr/channelBreakpoints.js 
     * Configurable object providing the values of the channel breakpoints.
     * Used by multichannel widgets indirectly via deliteful/features, and used directly by responsive widgets (ResponsiveColumns for now) that want to share the same default values of the breakpoints as multichannel widgets and to allow a global static configuration using `require.config()`.
* Modified Combobox and ResponsiveColumns to leverage the new infrastructure.
* Doc and tests.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -9,14 +9,14 @@ define([
 	"delite/HasDropDown",
 	"delite/keys",
 	"./list/List",
-	"./Combobox/ComboPopup",
+	"./features!desktop-like-channel?:./Combobox/ComboPopup",
 	"delite/handlebars!./Combobox/Combobox.html",
 	"requirejs-dplugins/i18n!./Combobox/nls/Combobox",
 	"delite/theme!./Combobox/themes/{{theme}}/Combobox.css"
 ], function (dcl, $, Filter, has, register, FormValueWidget, HasDropDown,
 		keys, List, ComboPopup, template, messages) {
 	/**
-	 * A form-aware and store-aware widget leveraging the `deliteful/list/List`
+	 * A form-aware and store-aware multichannel widget leveraging the `deliteful/list/List`
 	 * widget for rendering the options.
 	 * 
 	 * The corresponding custom tag is `<d-combobox>`.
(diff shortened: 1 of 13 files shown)


AdrianVasiliu:
Pushed via 26e4c5b3839c383a3f755b906f9dba2493c0093d.

PR #464, opened 2015-01-21
Switch: fixing disabled mode

tkrugg:
- removed any interaction with the widget when disabled
- changed the styling to differentiate disabled from enabled
- added a test that checks that clicking on the knob doesn't toggle the switch when disabled
 (I should probably add one that tests the sliding as well).

Styling the disabled mode for each theme.
image 
image 
image

build happening [here](https://travis-ci.org/tkrugg/deliteful/builds/47801380)

### DIFF
--- a/Switch.js
+++ b/Switch.js
@@ -75,50 +75,46 @@ define([
 		},
 
 		_pointerMoveHandler: function (e) {
-			if (! this.disabled) {
-				var dx = e.clientX - this._curX,
-					cs = window.getComputedStyle(this._pushNode),
-					w = parseInt(cs.width, 10);
-				if (!this._drag && Math.abs(e.clientX - this._startX) > 4) {
-					this._drag = true;
-					$(this._innerNode).removeClass("-d-switch-transition");
-					$(this._pushNode).removeClass("-d-switch-transition");
-					$(this._innerWrapperNode).removeClass("-d-switch-transition");
-				}
-				this._curX = e.clientX;
-				if (this._drag) {
-					// knobWidth and switchWidth are sometimes wrong if computed in 
-					// attachedCallback on Chrome so do it here
-					this._knobWidth = parseInt(window.getComputedStyle(this._knobNode).width, 10);
-					this._switchWidth = parseInt(window.getComputedStyle(this).width, 10);
-					var nw = this.isLeftToRight() ? w + dx : w - dx,
-						max = this.checked ? this._switchWidth : this._switchWidth - this._knobWidth,
-						min = this.checked ? this._knobWidth : 0;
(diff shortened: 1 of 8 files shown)


Switch.js
```diff
@@ -58,60 +58,67 @@ define([
...
 		_pointerMoveHandler: function (e) {
-			var dx = e.clientX - this._curX,
-				cs = window.getComputedStyle(this._pushNode),
-				w = parseInt(cs.width, 10);
-			if (!this._drag && Math.abs(e.clientX - this._startX) > 4) {
-				this._drag = true;
-				$(this._innerNode).removeClass("-d-switch-transition");
-				$(this._pushNode).removeClass("-d-switch-transition");
-				$(this._innerWrapperNode).removeClass("-d-switch-transition");
+			if (! this.disabled) {
```
cjolif: this is not strictly enforced but I think our policy is to not put spaces between `!` operator and the expression? (same at another place below).

AdrianVasiliu: +1

Switch.js
```diff
@@ -58,60 +58,67 @@ define([
...
-				this.checked = !this.checked;
-			} else {
-				this._drag = false;
-				var cs = parseInt(window.getComputedStyle(this._pushNode).width, 10);
-				var m = parseInt(window.getComputedStyle(this._pushNode).marginLeft, 10);
-				this.checked = cs + m + this._knobWidth / 2 >= this._switchWidth / 2;
-			}
-			if (this.checked !== oldCheckedValue) {
-				this.emit("change");
+			if (! this.disabled) {
```
cjolif: If for any reason disabled state is changed between point move and point up, the this._drag flag won't be reset to false.
Maybe we are trying to be too smart here and we could just support the disable state and the initialization of the interaction (i.e. on pointerDown).

tkrugg: I agree this isn't very solid.

But I doubt we want the give the user the ability to enable/disable the widget only on the start.
if you think of a form, you could have the state of switch depending of other inputs in the forms. It's a quite common use case I believe.

cjolif: Hmm... do you mean I meant we should only support it at widget initialization? Obviously not, I meant (and wrote) "at initialization of the interaction (i.e. on pointerDown)"

tkrugg: ah sorry. ok. 
Wouldn't it be better if we actually set/remove the listeners when enabled/disabled?

cjolif: If that's what you meant, I don't think having move listeners registered as soon as you enable the component is a good idea. You can get a lot of move events so a best practice is usually to register move listeners on when the interaction starts.

cjolif:
merged in 4126a33fcc8f28355bbb11a83a0c49f3127a8972

PR #472, opened 2015-01-23
Adapt doc layout to fix the navigation of ibm-js.github.io

clmath:
This is to fix the menu of ibm-js.github.io.

If you navigate to http://ibm-js.github.io/deliteful/docs/master/Button.html you can see that the `Docs` item in the menu is not highlighted.

### DIFF
--- a/docs/Button.md
+++ b/docs/Button.md
@@ -1,5 +1,5 @@
 ---
-layout: default
+layout: doc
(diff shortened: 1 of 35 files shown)


clmath:
Merged in 62979a7d21a0938f49fba8f7b8664a57781c51e8

Issue #376, opened 2014-10-30
Add iOS tests back in for deliteful.


edchat:
https://github.com/ibm-js/deliteful/issues/296 moved to Intern 2.1.1, so now we should be able to add the iOS tests back, but we were getting errors in some of the tests.

PR #475, opened 2015-01-23
Had to fix these tests or add some skips to pass the iOS tests.

edchat:
This is for https://github.com/ibm-js/deliteful/issues/376

### DIFF
--- a/tests/functional/Switch.js
+++ b/tests/functional/Switch.js
@@ -78,7 +78,7 @@ define([
 		"Switch key nav": function () {
 			this.timeout = intern.config.TEST_TIMEOUT;
 			var remote = this.remote;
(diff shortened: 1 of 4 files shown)


wkeese:
It would be better if there were tickets for the failing tests, or at least explanations of why you are skipping them.   "Skipping this test on iOS." just states what's already obvious.

AdrianVasiliu:
> or at least explanations of why you are skipping them

The comment exists already in most places - and could be copied in the places where the skipping has been added. The comment, for what it is worth, is:
"For some reason, the testing with click() doesn't pass on iOS for now, although it does seem to work when testing manually on the device. TODO: is it due to intern? other reason?"

PR #476, opened 2015-01-24
startup() calls: remove or replace w/attachedCallback()

wkeese:
Changes to doc, samples, and tests to remove unneeded startup() calls and convert other startup() calls to attachedCallback()

Reasoning: eventually we'll probably remove the startup() method, and instead:

1. either tell the users to call attachedCallback() instead of startup()
2. or perhaps we will set up listeners to call attachedCallback() automatically.

But for now we can avoid that whole issue by using Widget#placeAt() to position the node in the DOM tree.  placeAt() already calls attachedCallback(), thus there's no need for a manual call to startup().

### DIFF
--- a/ToasterMessage.js
+++ b/ToasterMessage.js
@@ -417,7 +417,7 @@ define(["dcl/dcl",
 			} else {
 				wrapper.appendChild(this);
 			}
-			this.startup();
+			this.attachedCallback();
 
 			// starting timer
 			if (this.isExpirable()) {
--- a/docs/Button.md
+++ b/docs/Button.md
@@ -40,7 +40,6 @@ See [`delite/Widget`](/delite/docs/master/Widget.md) for full details on how ins
   ], function (register, button) {
      var b = new Button({label: "I am a Button"});
      b.placeAt(document.body);
-     b.startup();
 });
 ```
 
--- a/docs/Checkbox.md
+++ b/docs/Checkbox.md
@@ -59,10 +59,8 @@ src="http://jsfiddle.net/ibmjs/7sxarg93/embedded/result,html,js">
 
      var cb = new Checkbox({checked:true});
      cb.placeAt(document.body);
-     cb.startup();
      cb = new Checkbox({disabled:true, name: "option1"});
      cb.placeAt(document.body);
-     cb.startup();
 });
 ```
 
--- a/docs/Combobox.md
+++ b/docs/Combobox.md
@@ -82,7 +82,6 @@ require(["delite/register", "dstore/Memory", "dstore/Trackable",
     // Create the Combobox
     var Combobox = new Combobox({list: list, selectionMode: "multiple"});
     Combobox.placeAt(document.body);
-    Combobox.startup();
 });
 ```
 
@@ -191,7 +190,6 @@ set for `value` on the `List` instance, for example:
     // Create the Combobox
     var combobox = new Combobox({list: list, ...});
     combobox.placeAt(document.body);
-    combobox.startup();
 ```
 
 or in markup:
--- a/docs/LinearLayout.md
+++ b/docs/LinearLayout.md
@@ -53,7 +53,6 @@ require(["deliteful/LinearLayout", "requirejs-domready/domReady!"], function (Li
   layout.addChild(centerChild);
   layout.addChild(rightChild);
   layout.placeAt(document.body);
-  layout.startup();
 });
 ```
 
--- a/docs/ProgressBar.md
+++ b/docs/ProgressBar.md
@@ -57,7 +57,6 @@ See [`delite/Widget`](/delite/docs/master/Widget.md) for full details on how ins
 
      var pb = new ProgressBar({max:100, value: 0});
      pb.placeAt(document.body);
-     pb.startup();
      pb.value = 20;
 });
 ```
--- a/docs/ProgressIndicator.md
+++ b/docs/ProgressIndicator.md
@@ -62,7 +62,6 @@ ProgressIndicator must be active to become visible and start its animation.
 
      var pi = new ProgressIndicator({active: true});
      pi.placeAt(document.body);
-     pi.startup();
 
      register.parse();
      //do some other tasks (load data...)
--- a/docs/RadioButton.md
+++ b/docs/RadioButton.md
@@ -55,13 +55,10 @@ allowfullscreen="allowfullscreen" frameborder="0">
 
      var sw = new RadioButton({checked:true, name: "category", value: "SUV"});
      sw.placeAt(document.body);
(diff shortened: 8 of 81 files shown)


cjolif:
> to remove unneeded startup() calls

Is that all of them or are there any remaining startup calls still in after this PR?

wkeese:
It's all of them.

AdrianVasiliu:
> But for now we can avoid that whole issue by using Widget#placeAt() to position the node in the DOM tree. placeAt() already calls attachedCallback(), thus there's no need for a manual call to startup().

This would imply that we promote the use of placeAt while we discourage (if not forbid) the use of DOM API (appendChild)? If we do support appendChild, not having samples/tests using appendChild (in conjunction with calls of startup() for now) wouldn't be a problem?

Also, placeAt doesn't not call startup() in all cases, see the TODO at
https://github.com/ibm-js/delite/blob/master/Widget.js#L387
(maybe related, there's also a TODO about attachedCallback at https://github.com/ibm-js/delite/blob/master/Widget.js#L383).

Finally, just curious, do all tests pass with this PR?

wkeese:
> If we do support appendChild, not having samples/tests using appendChild (in conjunction with calls of startup() for now) wouldn't be a problem?

Well, I want to get rid of the startup() method completely.

I could though change the patch to not change existing appendChild() calls, but just use attachedCallback() instead of startup().

> This would imply that we promote the use of placeAt while we discourage (if not forbid) the use of DOM API (appendChild)?

Yes and no.

Certainly not forbidding using appendChild(), but if you call appendChild() then afterwards you need to call attachedCallback(), and if you have nested widgets then you'd need to call attachedCallback() on those nested widgets too.

Since that's an unwieldy API I think it's better to promote using placeAt(), at least until if/when we set up listeners to make attachedCallback() get call automatically.   (A nicer API at the expense of performance.)

> Finally, just curious, do all tests pass with this PR?

Well, I think the tests are running as well as they were before the patch.  list-prog-5.html is failing with or without then PR, and there are intermittent failures like ViewStack Markup.

wkeese:
> Also, placeAt doesn't not call startup() in all cases

Yes, but it always called attachedCallback().

AdrianVasiliu:
> Yes, but it [placeAt] always called attachedCallback().

Except that it doesn't recurse into child widgets (https://github.com/ibm-js/delite/blob/master/Widget.js#L383), while we have widgets that create subwidgets, unless they call themselves attachedCallback or placeAt?

> I think the tests are running as well as they were before the patch.

ok

> list-prog-5.html is failing with or without then PR

AFAIK list-prog-5.html is not failling anymore since @cjolif recent fix.

wkeese:
> Except that [placeAt] doesn't recurse into child widgets

Right, so what I've done in wkeese/delite@034749d4931af530bd7c0c02c83de9089c15cd57 is to make placeAt() recurse into child widgets for descendants of Container.   There's also wkeese/delite@b1fd8036c989d6a8d5210ef4c1da93d36d6b3fd5 to handle widgets in templates.  I think that those two together are just enough to get the amount of recursion that we need, without overdoing it.

wkeese:
As per the meeting, I pushed these changes but left startup() working exactly as-is.

wkeese:
Oh, nevermind, I thought this was the delite PR.   This one is still open (and un-pushed).

edchat:
I thought the plan was to leave startup in delite as deprecated, but to go ahead and make these changes to avoid having deliteful use deprecated apis.

edchat:
Never mind, ignore my previous comment. cjolif explained what happened.

wkeese:
@AdrianVasiliu and @edchat - I changed this PR so that it just removes unneeded startup() calls, and replaces the other ones with attachedCallback().   In other words I rolled back the changes to convert appendChild() to placeAt().

Combobox, or at least the unit test, has more serious problems though, so I left that and filed #481 and #482.

@edchat - You were going to doublecheck the samples.   When you are finished, you can push the PR, or let me know to do it.

edchat:
I had problems with this PR (many list tests were failing on Chrome) without this additonal fix from Bill:
https://github.com/wkeese/delite/commit/75c8ff1a8fcb212ebdb963200cb6a3e65cf5fa89
But with that commit included all of the tests passed and I did not see any problems with the samples.

cjolif:
@edchat looks like you are pointing a commit in @wkeese fork, is that in main repo as well? (otherwise we are in trouble?)

edchat:
@cjolif, wkeese will have to merge that commit with this PR before he merges the PR.  If he merges this without that commit being merged we would be in trouble, but I wanted to let him know that for me the tests pass with that update.  wkeese had told me he would think about that update over night and decide in the morning if it should be merged.

wkeese:
OK, I pushed yet another beta of delite (0.6.0-beta.3) with that fix and the fix to use dpointer 0.5.0, then pushed this fix to deliteful.

The following test fails consistently (for `grunt test:remote:runner`) with and without this PR:

>> Test main - deliteful/features - functional - channel flags and breakpoint flags FAILED on safari 7.0.6 on MAC:
AssertionError: not checked with platform: MAC deviceName: null: expected false to be true

cc @AdrianVasiliu

AdrianVasiliu:
> The following test [deliteful/features] fails consistently (for `grunt test:remote:runner`) with and without this PR:

But in the travis build it didn't seem to be failing before the PR:
https://travis-ci.org/ibm-js/deliteful/builds/48534124
And a build after the PR is currently ongoing, without showing the failure so far.

AdrianVasiliu:
Apparently your `grunt test:remote:runner` leads to different UA for Safari on Mac OS X than the travis build, with "Mac" in the platform rather than "MAC OS X". I'll adjust the test to accept both.

AdrianVasiliu:
> I'll adjust the test to accept both ["MAC" and "OS X"]

Done via 04debcc093b7835a0adb614f87e8c852e1b7305e.
(although https://docs.saucelabs.com/reference/platforms-configurator mentions `OS X ...` for the Mac platform on all OS X versions, anyway, this change should make the test more robust).

wkeese:
OK, thanks, it's passing for me now.  It's bizarre how `remote.environmentType.platform` changes.  Perhaps something about which saucelabs account is being used.

AdrianVasiliu:
> OK, thanks, it's passing for me now. It's bizarre how remote.environmentType.platform changes. 

I confirm, very bizarre... It's passing for you, but it didn't pass in the travis build, because it gets "MAC" platform when running on iOS... (platform: MAC deviceName: IPHONE SIMULATOR browserName: IOS). I've re-fixed it via 2f4a2239f48a5d8a6da8d0640459a2e86fdad2ad...

PR #483, opened 2015-01-27
List.js: pass event to KeyNav's navigation API

AdrianVasiliu:
This small change in List is the deliteful counterpart of https://github.com/ibm-js/delite/pull/372 (which is now merged).
This is to be merged AFTER delite will be re-released to include the change in delite/KeyNav.

### DIFF
--- a/list/List.js
+++ b/list/List.js
@@ -1044,7 +1044,7 @@ define([
 		},
 
 		// Simple arrow key support.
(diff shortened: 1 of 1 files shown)


AdrianVasiliu:
Pushed via aade19e3ae73f14adbade397df7290c7d42630ef.

PR #491, opened 2015-01-28
ProgressBar CSS improvements

dmandrioli:
- Add a LESS variable @height for consistency with @line-height. These two variables control the default height of the widget and the default vertical position of the text
 - Ensure inner font size inherit the widget font-size
 - Add some comments, remove a "overflow:hidden" not needed

### DIFF
--- a/ProgressBar/themes/bootstrap/variables.less
+++ b/ProgressBar/themes/bootstrap/variables.less
@@ -1,2 +1,3 @@
-@transition-params: 0.3s linear 0s;
-@line-height: 1.6em;
+@height: 1.8em; // Controls the height of the widget (180% of the font-size)
+@line-height: 1.6em; // Controls the vertical position of the text
+@transition-params: 0.3s linear 0s; // Controls the animation
--- a/ProgressBar/themes/ProgressBar_template.less
+++ b/ProgressBar/themes/ProgressBar_template.less
@@ -1,6 +1,6 @@
 .d-progress-bar {
   display: inline-block;
-  position: relative;
+  position: relative; // Needed for absolutely positioned children
   padding: 0;
   width: 100%;
   .d-progress-bar-styles;
@@ -28,7 +28,7 @@
   width: 100%;
   margin: 0;
   padding: 0;
-  overflow: hidden;
+  overflow: hidden; // Needed for the alignment of the two parts of the text
   .d-progress-bar-msg-styles;
 }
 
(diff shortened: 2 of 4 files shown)


ProgressBar/themes/bootstrap/variables.less
```diff
@@ -1,2 +1,3 @@
-@transition-params: 0.3s linear 0s;
-@line-height: 1.6em;
\ No newline at end of file
+@height: 1.8em; // Controls the height of the widget (180% of the font-size)
```
cjolif: isn't that "dangerous" to have such non-scoped LESS (i.e. not pb-height for example) variable naming just for a particular widget? What is our practice elsewhere? Can this collapse with other widget defining a similar variable?

dmandrioli: Not sure if it's dangerous or not, probably not, but there is definitely something to do on LESS consistency across widgets.
There are variables such as @star-icon-size, others like @d-switch-w or too short names like @height.

Where are defined these LESS variables is also not consistent:
- In a variables.less for the Switch in `Switch/variables.less` which is common to all themes.
- in the theme LESS file for others

About naming, I propose to follow the pattern of @d-switch-xxx for all widgets.
About where to define LESS variables, having a variables.less for each widget could be interesting in the perspective of a theming tool, to facilitate a kind of automatic introspection mechanism of all LESS variables. But I guess it's not the time to do that and we should remove the only variables.less of Switch.

I would prefer to refactor all LESS variables in a unique commit after merging all CSS related PRs.

cjolif: > I would prefer to refactor all LESS variables in a unique commit after merging all CSS related PRs.

Ok. Can you create an issue for that? (yes one more...)

dmandrioli: > Ok. Can you create an issue for that? (yes one more...)
Sure, https://github.com/ibm-js/deliteful/issues/496.

AdrianVasiliu: > Can this collapse with other widget defining a similar variable?

On a related note, shouldn't we systematically apply a naming pattern (with the "d-" prefix) also for CSS classes? I'm not sure whether we have other cases, the only one I have in mind is the `fill` class (https://github.com/ibm-js/deliteful/blob/0.6.0-alpha/LinearLayout/themes/LinearLayout_template.less#L12). I can create an issue for it.

dmandrioli: We also have https://github.com/ibm-js/deliteful/blob/master/docs/Button.md#styling-css-classes
Feel free to create an issue.

dmandrioli:
Closed by d984dc8914b7844659a81ebc80ca45dfaa984453

PR #502, opened 2015-01-30
Combobox: properly separate "keynav-child-navigated" and click handlers.

wkeese:
Also reduce a bunch of code duplication.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -385,67 +385,36 @@ define([
 			this._initValue();
 		},
 		
-		/**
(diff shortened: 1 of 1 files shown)


AdrianVasiliu:
Thanks @wkeese. Pushed via 36ec767a598157a61ada2566c4feee24e8dd42a8.

Also pushed this fix: 7a85aaccd27d2b9efec4d40b94ddd774650d6236 to avoid closing the popup when clicking a list category item, which was a regression I previously introduced (thus removed the TODO from your PR).

Issue #467, opened 2015-01-22
Bidi: tests are excluded because causing trouble, fix issues and re-enable theme


cjolif:

cjolif:
See also https://github.com/ibm-js/deliteful/pull/451#discussion_r23389604

semion1956:
> Thinking more about this, I think the only viable solution will be to have bidi tests totally separated with its own intern config (inheriting from the standard one) and another grunt target to launch bidi tests (and maybe yet another one to run both bidi and non bidi).

Can it be the same grunt task with additional bidi subtasks (like local.bidi, local.android.bidi etc - not sure that I use the correct term), configured by intern.bidi, which sets bidi flag and refer to bidi suites only?

clmath:
There is no such things as subtasks in grunt. The "local.android" or "local.ios" are just regular grunt task target and the "dot notation" is simply a convention. You could indeed create new "bidi" targets on the same model (like "local.bidi" etc.)

PR #504, opened 2015-02-02
Add bidi targets for grunt intern

semion1956:
Issue #467


wkeese:
I think you also should:

1. modify travis.yml to run both "grunt test:remote" and "grunt test:remote.bidi"
2. remove the commented out line from unit/all.js and functional/all.js loading "bidi/all"

Also, ITSM your intern config files could set the bidi flag to true by:

```js
intern.loader.config["requirejs-dplugins/has"].bidi = true;
```

Maybe that's better than what you have now, because currently you have 8 lines of code that overwrite the whole configuration object.

I also noticed (unrelated to your PR) that the current intern config files have some redundancy:  intern.local.android and intern.local.ios should be "extending" from intern.local rather than from intern.

AdrianVasiliu:
> modify travis.yml to run both "grunt test:remote" and "grunt test:remote.bidi"

The deliteful tests currently fail randomly with timeout in travis because they last for a duration not far from the 50 min. max. Maybe there are ways to overcome this limit, or to speed up our tests, but if with the addition of bidi we largely increase the duration, for now we may be in trouble. Or maybe the bidi run would be distinct, with a separate timeout?

cjolif:
I think @AdrianVasiliu is right that we should be careful not adding too much to our travis build because of that 50 minutes limited. That said maybe a solution is to use a travis build matrix.

See: http://docs.travis-ci.com/user/speeding-up-the-build/

We would have two environments bidi and non bidi that would run in parallel. 

Now in practice the parallelism might not work because of SauceLabs which will give us only 3 VMs at a time. Might be worth trying however.

semion1956:
So, what is the final decision? 
If problem still in the different configurations, required from loader, can we simply test subclasses, created specifically for this case (like some 'BidiButton', which inherits from deliteful/Button, delite/Bidi and deliteful/Button/bidi/Button)?

PS What happened with this PR? Why so many commits are shown here?

cjolif:
> So, what is the final decision? 

If you have time to test the build matrix we can go for it. Otherwise let's put bidi tests out of the CI build until we have a solution.

> PS What happened with this PR? Why so many commits are shown here?

We have cleaned up  deliteful history. So you need to update your PR accordingly. This was mentionned on delite-dev mailing list.

lbod:
@cjolif @AdrianVasiliu re timeouts, have we tried setting up caching yet? 

e.g. http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/ and from the page christophe linked http://docs.travis-ci.com/user/caching/
node_modules should definitely be cached and there's probably a bunch of other stuff too

I can try out over the next few days to see the differences if you haven't done already

cjolif:
@lbod we have not, so feel free to test. That said unfortunately I don't think the npm installation phase is significant in terms of overall % running time.

lbod:
@cjolif I'd think you're right, it's a one time install but some analysis of where/why tests are taking so long might be something I could look at. Other areas to look at would be Saucelabs https://support.saucelabs.com/entries/23388772-Tips-for-lean-speedy-tests 
Also it looks like the caching support in travis is limited to private repos.
When I watch a suite in action, it seems to take an age before the first test is actually run

wkeese:
> have we tried setting up caching yet?

I wonder if ibm-js's CSS and JS is getting cached, or if every functional test reloads all those modules (for example List.js).

> When I watch a suite in action, it seems to take an age before the first test is actually run

I'm guessing that's from all of Intern's (i.e. Leadfoot's) feature testing.

wkeese:
I rebased your code and checked in a modified version of your PR as 15cc63d2473b5f478f042dc39cf5a16a1dfdeff8.

PR #505, opened 2015-02-03
Fixed Switch high contrast mode, fixes #477

### LINKED ISSUE #477: Switch: High-contrast mode should be supported
Rendering in high contrast mode (IE and FF) is not correct:
image

tkrugg:
- Added a transparent border that will show in high contrast mode while
still invisible in normal mode.
- fixing checkbox sample title
- added a standelone sample for switch

Here's how it looks now in high contrast mode. (Tested under firefox and ie)
image

### DIFF
--- a/Switch/themes/Switch_template.less
+++ b/Switch/themes/Switch_template.less
@@ -17,7 +17,12 @@
 }
 
 // internal classes
-
+.-d-switch-block {
+	display: inline-block;
+	position: absolute;
+	top:0;
+	border-width: 0px;
+}
 .-d-switch-inner {
 	position:absolute;
 	width: @d-switch-knob-w;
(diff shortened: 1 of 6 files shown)


wkeese:
The PR seems reasonable to me.

A further improvement might be to set `bottom: -1px` on the knob too, so that the knob's height will stay matched to the `<d-switch>` nodes if someone modifies it's height.   Of course that won't update the knob's width, which is currently hardcoded to 30px.

tkrugg:
@wkeese I can't set `bottom` because the parent has no height defined.

wkeese:
> I can't set bottom on the knob because its parent has no height defined.

Hmm, strange that you can set `top` but not `bottom`, but OK.

tkrugg:
Sorry I meant I *can* set `bottom`, and it *will* change the height, but it'll change it to a value that depends on the parent's and won't produce the result we want, see http://jsfiddle.net/tkrugg/z5mt42La/

cjolif:
merged in ae1f312ec2b1d77a55cf84808d8225d52d43966b

Issue #442, opened 2015-01-06
Checkbox and RadioButton labeling should be easier


dmandrioli:
If you just drop some checkboxes and radiobuttons with default configuration, you get something like
![image](https://cloud.githubusercontent.com/assets/1109677/5631339/175c4348-95c6-11e4-9594-47378c194da7.png)
Source: http://jsfiddle.net/ibmjs/c3kgt98w/

Labels alignment is weird and different between the two widgets:
 -  Checkbox label is too high, the text baseline should be aligned with the bottom of the square
 - Radio label is too low
 - Checkboxes and Radiobuttons should be aligned

As an example, here is what you get using Bootstrap
![image](https://cloud.githubusercontent.com/assets/1109677/5631445/e78a623e-95c6-11e4-83ba-97e79d493e8d.png)

tkrugg:
I've been working on this, as part of issues https://github.com/ibm-js/deliteful/issues/299#issuecomment-64398124 and https://github.com/ibm-js/deliteful/pull/415

I fixed the outline, but also the alignement on radio buttons.
i'm about to send a [PR](https://github.com/tkrugg/deliteful/compare/RadioButton/fixing-style) but I need to work on it some more, even maybe include the fix for checkbox.

AdrianVasiliu:
Just to mention that adding vertical-align: middle to d-radio-button gives this result with your jsfiddle:

![image](https://cloud.githubusercontent.com/assets/1583981/5632426/eb7f7c0a-95ce-11e4-9afb-57c1d67f7a68.png)

wkeese:
Also fixed by 4dfd44b8f5bb91ef27a66ddff4f688c7135ca1f3.

Issue #489, opened 2015-01-27
RadioButton: interoperability with bootstrap.


tkrugg:
- [ ] When wrapped inside a class=radio element, like you would do it with a regular input[type=radio], deliteful/RadioButton shows two focus outlines.
  Hopefully, this will disappear with https://github.com/ibm-js/deliteful/issues/442#issuecomment-68891377

``` html
<div class="radio">
    <label>
        <d-radio-button></d-radio-button> Option 2
    </label>
</div>
```
- [ ] Outline color should follow the validation state style.

``` html
<div class="has-success radio">
    <label>
        <d-radio-button></d-radio-button> Option 2
    </label>
</div>
```
- [ ] alignment is broken: already filed in issue https://github.com/ibm-js/deliteful/issues/442

cjolif:
This issue is mostyle about the 2nd point then. Other issues are dealt as part of #442.

wkeese:
Closing this as we have no plans anymore to support interoperability with bootstrap.

Issue #488, opened 2015-01-27
Checkbox: interoperability with bootstrap.


tkrugg:
The checkbox is simply invisible when wrapped inside a div.checkbox like you'd usually do it with input[type=checkbox].

wkeese:
Closing this as we have no plans anymore to support interoperability with bootstrap.

Issue #299, opened 2014-09-17
deliteful/Switch: on desktop, clicking it adds an unpleasant outline


AdrianVasiliu:
At least as it can be seen in samples/Buttons.htmln on desktop browsers clicking the Switch adds an outline to it:

![image](https://cloud.githubusercontent.com/assets/1583981/4301222/b9aeafd6-3e4c-11e4-8f59-0cb4f19fe988.png)

This is the same focus outline as when navigating via tab - due to `d-switch.d-focused` CSS rule - but I think should be avoided when simply clicking. For comparison, no such outline with dojox/mobile/Switch (while it does have the outline when tab-navigating).

Reproduced on Win7 with Chrome 37, FF 32, IE10.

PR #508, opened 2015-02-03
Fixing RadioButton and Checkbox styles, solves #442 #299 #415 #489 #488

tkrugg:
- fixed labels positionning, solves #442
- fixed inconsistent outlines, solves #299 and #415
- fixed interop with bootstrap, solves #489 and #488
  - fixed positioning inside .checkbox{,-inline} and .radio{,-inline}
  - added validation state dependant focus
  - using bootstrap variables for sizing, in the bootstrap theme

NB: This is **not ready for merge**. It's only here to gather some feedback.before going further.

Additionally:
- for now, removed holodark and ios themes for Checkbox -- this is a step back for checkbox, not radio button which didn't have any in the first place. It's not hard to do, I just need to know what priority it is.
- moved radiobutton and checkbox to svg. This definitively solves the positioning issue we were experiencing on the various browsers, while combining DOM and characters.
- not handling rtl for now -- prior version of d-checkbox did
- high contrast mode looks "ok" (tested on IE and FF), though it's not perfect. In high contrast mode, a native checkbox would have a white checkmark inside a black square. The HC restyling doesn't affect svg.
- radio-button and checkbox's css are almost a 100% identical. Any ideas on how to avoid code duplication in a clean way?
- the bootstrap theme stylesheet contains references to bootstrap specific classes to help with interop (see [here](https://github.com/tkrugg/deliteful/blob/Checkbox-Radiobutton/fixing-style%23442/Checkbox/themes/bootstrap/Checkbox.css#L5-L9) and [here](https://github.com/tkrugg/deliteful/blob/Checkbox-Radiobutton/fixing-style%23442/Checkbox/themes/bootstrap/Checkbox.css#L71) )


Checkbox/Checkbox.html
```diff
@@ -1,4 +1,10 @@
-<template class="d-reset d-inline" role="presentation">
-	<input type="checkbox" role="checkbox" attach-point="focusNode,valueNode"
-		checked="{{checked}}" value="{{value}}" name="{{name}}"/>
+<template role="presentation">
+
+	<input class="d-reset" type="checkbox" attach-point="focusNode,valueNode"
```
dmandrioli: `d-reset` set font, line-height and color. This looks not needed here if only padding and margin must be set. Same comment for radiobutton.

Checkbox/themes/Checkbox_template.less
```diff
@@ -1,56 +1,55 @@
...
 .d-checkbox {
 	position: relative;
-	.d-checkbox-styles();
-	&.d-focused { // from tab-focus()
-		.d-checkbox-focused();
+	display: inline-block;
+	/* #408 Enforce content-box to ensure correctly centered check mark */
+	/* even if border-box is used everywhere else for instance */
+	/* via delite/themes/defaultapp.css */
+	box-sizing: content-box;
```
dmandrioli: Enforcing box-sizing looks obsolete with this refactoring. Same comment for radiobutton.

Checkbox/themes/Checkbox_template.less
```diff
@@ -1,56 +1,55 @@
...
+	
+	&.d-disabled *, fieldset[disabled] & * {
+		cursor: not-allowed;
 	}
-	&::before {
-		overflow: hidden;
+
+	&-visual {
+		overflow: visible;
```
dmandrioli: I noticed that overflow visible make the border stronger. But I don't understand how it works, could you comment?
The border of the radiobutton is lighter with the same directive.
I wonder if removing overflow on both would solve the question: more consistency between checkbox and radiobutton and less CSS.

dmandrioli:
The radio button is a bit smaller than the checkbox. Adjusting the radio button size could also improve aligment with label:
image

wkeese:
The SVG fixes #442?   That's nice, but IIUC the downside is that it breaks theme customizatio, i.e. you can't have different checkbox icons depending on theme, although perhaps you can control the color.

Checkbox/Checkbox.html
tkrugg: oversight on my part.

tkrugg:
> The SVG fixes #442? 

SVG is fixing another positioning problem (reported in https://github.com/ibm-js/deliteful/issues/408, which I forgot to mention)
- position of the checkmark inside the square
- position of the dark circle inside the bigger light circle in the radio-button

When the checkbox was using DOM + a unicode character as a checkmark, it could be easily affected by a lot of CSS properties such as line-height, font-size, margins, padding... It was hard to make it look the same cross-browser (1px makes a huge difference at theses sizes) and was hurting the interoperability with bootstrap. 

Radiobutton is apprently much simpler an could be done only with DOM, but I wasn't satisfied with the way it looked at small sizes: the version below was implemented with DOM only in chrome.

image

Part of my work on d-icon is see how we can have a different icon pack for each theme. This should solve your problem.

Checkbox/themes/Checkbox_template.less
tkrugg: <svg> has by default overflow:hidden. At some sizes it hides the borders of the graphic inside or thinners it.

Checkbox/themes/Checkbox_template.less
tkrugg: true. I'll remove it

tkrugg:
> The radio button is a bit smaller than the checkbox. 

In which browser do you see it? CSS wise they really are the [same](https://github.com/tkrugg/deliteful/blob/Checkbox-Radiobutton/fixing-style%23442/Checkbox/themes/bootstrap/Checkbox.css#L38-L41) size in [px](https://github.com/tkrugg/deliteful/blob/135e168a5fda2c93ae372136a0598c1da96e13a0/RadioButton/themes/bootstrap/RadioButton.css#L35-L38).

dmandrioli:
> In which browser do you see it? CSS wise they really are the same size in px.

In Chrome: 
image

and IE11:

image

As mentioned, this is probably related to the different behaviour of overflow:visible on checkbox and radiobutton.

wkeese:
> SVG is fixing another positioning problem (reported in #408, which I forgot to mention)
> ...
> Part of my work on d-icon is see how we can have a different icon pack for each theme. This should solve your problem.

Well, IIUC you are researching SVG, and you are a big fan of SVG, but we haven't actually decided to switch from font-icons to SVG.  And even if we decide to switch, the end result won't look like what you have in this PR: you've hardcoded the icon into the widget's template, whereas ideally it should be part of the theme.

I don't have a strong opinion, but not sure it makes sense to make this change now.

tkrugg:
Every single thing you said is true. i'm just exploring here... never meant to merge this, like dazzlingly stated in bold first post.

If we ever decide to make the switch, the end result *will* look the same though, the markup will change, but no reason the stylesheet will. Strong opinion or not, please comment with it (see https://github.com/ibm-js/deliteful/issues/434).

@dmandrioli I looked at it. I found that it's the border-radius set on the svg element that hiding part of the content, it'll be fixed.

wkeese:
Ah right, sorry, you did write that.  Well anyway, it's a good proof-of-concept of using SVG for these icons.

I do think it would be easy to fix the alignment issues while still using font-icons, because it shouldn't be difficult to make one character align with other characters.  Actually, that should happen naturally, assuming we aren't adding CSS that messes it up.

wkeese:
I guess this PR got forgotten.  #442 #299 #415 are already fixed.  #489 #488 are still open (or perhaps the problem is fixed but the tickets weren't closed).

Issue #290, opened 2014-09-12
Button: extend delite/Container?


wkeese:
Button should probably extend delite/Container, so that Button widgets and subclasses can display complex app-defined markup, for example:

``` xml
<d-button>
   <d-progress-indicator></d-progress-indicator>
   refreshing...
</d-button>
```

Currently the textContent is copied by hand in preCreate:

``` js
createdCallback: function () {
    // Get label from innerHTML, and then clear it since we are to put the label in a <span>
    if (!this.label) {
        this.label = this.textContent.trim();
        this.innerHTML = "";
    }
},
```

Note that the `if (!this.label) {` condition is not very useful since user-defined properties have not yet been processed.

PR #513, opened 2015-02-04
Made Button a container, solves #290

tkrugg:
- made Button a Container
- removed the part that trims innerText to fill in `label`
- set containerNode before the icon and the label
- I updated the samples (they were setting label using innerText) and added an example with a ProgressIndicator inside a ToggleButton.
- updated doc

> Quoted from doc: The only string displayed by the button is the one defined by its `label` property.

I took the liberty of removing this part of the doc, i think it's no longer relevant.

### DIFF
--- a/Button.js
+++ b/Button.js
@@ -3,11 +3,11 @@ define([
 	"dcl/dcl",
 	"requirejs-dplugins/has",
 	"delite/register",
-	"delite/Widget",
+	"delite/Container",
 	"requirejs-dplugins/has!bidi?./Button/bidi/Button",
 	"delite/handlebars!./Button/Button.html",
 	"delite/theme!./Button/themes/{{theme}}/Button.css"
-], function (dcl, has, register, Widget, BidiButton, template) {
+], function (dcl, has, register, Container, BidiButton, template) {
 
 	/**
 	 * A Non-templated form-aware button widget.
@@ -23,9 +23,9 @@ define([
 	 * </style>
 	 * <button is="d-button" iconClass="iconForButton">Click Me</button>
 	 * @class module:deliteful/Button
-	 * @augments module:delite/Widget
+	 * @augments module:delite/Container
 	 */
-	var Button = dcl(Widget, /** @lends module:deliteful/Button# */ {
+	var Button = dcl(Container, /** @lends module:deliteful/Button# */ {
 
 		/**
 		 * The text to display in the button.
@@ -49,14 +49,6 @@ define([
 		baseClass: "d-button",
 
 		template: template,
-
-		preRender: function () {
-			// Get label from innerHTML, and then clear it since we are to put the label in a <span>
-			if (!this.label) {
-				this.label = this.textContent.trim();
-				this.innerHTML = "";
-			}
-		},
 
 		computeProperties: function (props) {
 			if ("title" in props || "label" in props) {
--- a/Button/Button.html
+++ b/Button/Button.html
@@ -1,4 +1,5 @@
 <template attach-point="focusNode">
-	<span attach-point="iconNode" class="d-icon {{this.iconClass ? this.iconClass : 'd-hidden'}}"></span
+	<span attach-point="containerNode"></span
+	><span attach-point="iconNode" class="d-icon {{this.iconClass ? this.iconClass : 'd-hidden'}}"></span		
 	><span attach-point="labelNode" d-hidden="{{!this.label}}">{{label}}</span>
 </template>
--- a/ToggleButton/ToggleButton.html
+++ b/ToggleButton/ToggleButton.html
@@ -1,4 +1,5 @@
 <template aria-pressed="{{checked}}" attach-point="focusNode" on-click="{{toggle}}">
-	<span attach-point="iconNode" class="d-icon {{this.checked &amp;&amp; this.checkedIconClass ? this.checkedIconClass : (this.iconClass ? this.iconClass : 'd-hidden')}}"
+	<span attach-point="containerNode"></span
+    ><span attach-point="iconNode" class="d-icon {{this.checked &amp;&amp; this.checkedIconClass ? this.checkedIconClass : (this.iconClass ? this.iconClass : 'd-hidden')}}"
 	></span><span attach-point="labelNode" d-hidden="{{!this.checkedLabel &amp;&amp; !this.label}}">{{this.checked &amp;&amp; this.checkedLabel ? this.checkedLabel : this.label}}</span>
 </template>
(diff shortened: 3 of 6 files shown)


wkeese:
Hmm, this doesn't seem right.   You have two separate nodes: `containerNode` and `labelNode`.  Shouldn't it just be one node?   Because, for example, `myButton.label = "OK"` should clear the button text set originally in markup, ex: `<d-button>hello</d-button>`.

tkrugg:
`.label` doesn't seem important in Button but it needs to be kept because it's used in ToggleButton (used along with `.checkedLabel`), and there are situations where we might want to have both, as to make the following kind of refresh button, embedding a ProgressIndicator:

image

Also currently, it's impossible to get an `<icon><text>` kind of button without a label.

wkeese:
> there are situations where we might want to have both, as to make the following kind of refresh button, embedding a ProgressIndicator

If a user wants a button with a label and a ProgressIndicator, he can easily do:

```html
<d-button>
   <d-progress-indicator></d-progress-indicator>
   refreshing...
</d-button>
```


> Also currently, it's impossible to get an <icon><text> kind of button with containerNode only.

That's not true, just check out http://download.dojotoolkit.org/release-1.10.2/dojo-release-1.10.2/dijit/tests/form/test_Button.html.  It has many examples like that.

cjolif:
What is missing is enhancing existing tests to tests for that new feature?

cjolif:
I personally have no issue having the label not clearing the content. Actually I think it is probably easier for the user to understand, there will be not trick of setting a particular property influencing the behavior.

tkrugg:
> That's not true, just check out http://download.dojotoolkit.org/release-1.10.2/dojo-release-1.10.2/dijit/tests/form/test_Button.html. It has many examples like that.

By "currently" I meant in this PR. But with the dijit samples I understand better the bahviour you described in https://github.com/ibm-js/deliteful/pull/513#issuecomment-72966489. I'll implement that.

wkeese:
> I personally have no issue having the label not clearing the content. Actually I think it is probably easier for the user to understand, there will be not trick of setting a particular property influencing the behavior.

Well, I think the behavior would be counterintuitive to users.  The common use cases of buttons are:

```html
<d-button id=myButton>OK</d-button>
```

or creating programatically:

```js
myButton = new Button({label: "OK});
```

99% of users (especially HTML programmers) will see these two as identical, using the terms "content", "text" and "label" interchangeably.   In other words, if the application later does:

```js
myButton.label = "Cancel";
```

Then I think users will expect the text displayed by the button to be "Cancel", not "OK Cancel" or "Cancel OK".

cjolif:
What you say seems to imply that label property must be treated differently than other properties so that it is easier for users. What I'm saying is that users find thing easier when they are consistent (i.e. in this case label is no different than other properties). I guess both are valid ways of seeing it. I'm a bit frighten that you find the first one more obvious because we are used to do that in Dijit. But that does not necessarily mean all users would see it like that.

edchat:
cjolif, the first example I see when looking up html5 button is:
 `<button type="button">Click Me!</button> ` so it seems like that is common.

cjolif:
@edchat so what, this will work?

edchat:
@cjolif when you said "I'm a bit frighten that you find the first one more obvious because we are used to do that in Dijit." I thought you meant that syntax.  Is that not the first one you were talking about?

cjolif:
> when you said "I'm a bit frighten that you find the first one more obvious because we are used to do that in Dijit." I thought you meant that syntax. Is that not the first one you were talking about?

I meant that `<button label="foo">bar</button>` having label "foo" overriding "bar" being dijit behavior (first solution) comparing to @tkrugg initial PR which was displaying both (2nd solution).

wkeese:
As @edchat suggested, I'm trying to match the behavior on native HTML, not trying to match dijit.   Also, `<button label="foo">bar</button>` is a weird example because no one would ever do that.

cjolif:
> As @edchat suggested, I'm trying to match the behavior of native HTML, not trying to match dijit.

I'm probably missing something but what in @tkrugg solution is not matching native HTML behavior? If you don't put a label property (which the native element does not have) you get the same visual rendering as the native button, aren't you? The only case where you have a different behavior is if you use a label property, but that's expected as anyway native button does not have a label property...

I guess I'll come back to my initial point which is why making label such a specific case? We have not other property fo the widget that is filled automatically by its content? Also if we do that we should also make sure to listen to changes of the content to re-apply it correctly?  (which it is not doing today?)

In other words doing that 100% correctly (or even 95%) is it seems to me quite some code and I don't really see the necessity of "paying" for it? We are not losing functionality without it? (i.e. you want to use label, you use it, you want to use child content you use it, but what is the thing we are losing by not replicating child content into label?)

wkeese:
> what is the thing we are losing by not replicating child content into label?

As I said above, having both "label" and "content" as distinct properties leads to unexpected behavior and makes it slightly difficult for the application to programatically change the text of a button that was created declaratively.

cjolif:
> makes it slightly difficult for the application to programatically change the text of a button that was created declaratively.

There are two ways to setup the button programmatically, if you actually set up a label property it should not be hard to change its label? 

You are right in the case you don't use the label property but the "content". But in that case you are not worst than a pure HTML button, or are you? If you are then I agree this is problematic, if not one more time I'm ready to live with this for the sake of simpler code.

wkeese:
> But in that case you are not worst than a pure HTML button, or are you?

With a vanilla HTML button like `<button id=myButton>OK</button>`, you can change the (declaratively or programatically set) label by setting `myButton.textContent` or `myButton.innerHTML`.   So, it's easy to do programatically, and perhaps more importantly, presumably works with MVC frameworks.

Unfortunately, this strategy doesn't work with a deliteful Button widget because the label is in a sub-node, not the root node.   You could tell the application to set `myButton.containerNode.textContent`, but I don't think that will work well with MVC frameworks.

Ideally we would support setting `textContent` or `innerHTML` directly but it seems too difficult to implement.  Thus, mapping `label` to `containerNode.textContent` seems like a good compromise.

wkeese:
> I guess I'll come back to my initial point which is why making label such a specific case? We have not other property fo the widget that is filled automatically by its content? 

Actually, we have many cases like that:

* Button and subclasses currently work like that (initial textContent --> `label` property)
* delite/Store subclasses (initial textContent --> `source` property)
* Combobox (initial child `<d-list>` node --> `list` property)

Regardless, the point of #290 is not to change existing behavior, but to enhance Button to support markup like:

```xml
<d-button>
   <d-progress-indicator></d-progress-indicator>
   refreshing...
</d-button>
```

If there's a consensus to change the existing behavior, we could do that separately.

PR #521, opened 2015-02-19
Bidi fixes and cleanup

### LINKED ISSUE #462: bidi support dependencies/flag
After some investigation it happens we have some "bidi"-type features under bidi flag and some are not. We need to decide either to make 2 clear categories one under flag, one without flag or to come up with a single category which is managed in a consistent manner.

We should also decide what the proper naming for the flag is.

-----
Initial issue:

StarRating is doing those requires:

```js
	"requirejs-dplugins/has!bidi?./StarRating/bidi/StarRating",
	"requirejs-dplugins/has!bidi?delite/theme!./StarRating/themes/{{theme}}/StarRating_rtl.css"
```

It would probably better better to centralize all the bidi stuff in one module (here StarRating/bidi/StarRating) and thus require the CSS from this module instead of from the regular StarRating module.

### LINKED ISSUE #517: sloppy usage of d-rtl class can make LTR widgets show up as RTL
There are a number of files that have CSS selectors just referencing d-rtl, for example:

```css
.d-rtl .d-progress-bar-indicator {
  .d-progress-bar-indicator-rtl-styles;
}
```

This selector will match too often, specifically in a case where the wrapper is RTL but the widget itself is LTR, maybe something like:

```html
<d-view-stack dir=rtl>
    ...
        <d-progress-bar dir=ltr>
```

The selector should reference the widget class too:

```css
.d-progress-bar.d-rtl .d-progress-bar-indicator {
  .d-progress-bar-indicator-rtl-styles;
}
```

wkeese:
Make basic RTL support work w/out `has("bidi")` flag, as suggested in #462.
Fixes were to List, ProgressBar, StarRating, SidePane, to roll CSS into main .less files and also, for StarRating, rolling JS code into main JS file.   The RTL CSS and code for these widgets is tiny, except for SidePane.

Also, fix naked uses of `.d-rtl` to be `.d-widget-name.d-rtl`.

Plus, fix some CSS typos.

I found that putting the RTL and LTR CSS together makes it easier to keep things in sync and notice bugs.

Fixes #462, fixes #517.

After this PR I'll update deliteful to run against latest delite, where isLeftToRight() was replaced by effectiveDir.


wkeese:
Squashed and pushed as 86ec1850b6f15e53f459d10bf420ca50603ea7f7.

PR #522, opened 2015-02-19
List: don't hardcode names of baseClasses for renderers

### LINKED ISSUE #268: List: stop using baseClass in get***() methods
As mentioned in https://github.com/ibm-js/delite/issues/271#issuecomment-54745268, it would be better for List to use the renderers' tag names `this.itemRenderer.tag` and `this.categoryRenderer.tag` rather than the baseClasses listed in `this._cssClasses.item` and `this._cssClasses.category`.

The main reason for this is actually to avoid synchronization issues.   Currently if a List subclass overrides the `itemRenderer` property, it also needs to update `this._cssClasses.item`, and likewise for categoryRenderer.

On a semi-related note, getItemRendererByIndex() could be further improved.  It currently fetches all the list items and then selects the n'th one, but you should be able to do a query like below to get the n'th one directly:

```
this.containerNode.querySelector("d-list-item-renderer:nth-of-type(5)")
```

wkeese:
Don't hardcode names of baseClasses for item renderer and category renderer
since they are already available through properties.  Fixes #268.

### DIFF
--- a/list/CategoryRenderer.js
+++ b/list/CategoryRenderer.js
@@ -14,9 +14,7 @@ define([
 	var CategoryRenderer = dcl(Renderer, /** @lends module:deliteful/list/CategoryRenderer# */ {
 
 		/**
-		 * CSS class of a category renderer. This value is expected by the
-		 * {@link module:deliteful/list/List deliteful/list/List widget}
-		 * so it must not be changed.
+		 * CSS class of a category renderer.
 		 * @member {string}
 		 * @protected
 		 */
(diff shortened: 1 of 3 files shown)


cjolif:
Looks good to me. Isn't the developer guide also mentioning this? If yes maybe it needs to be updated accordingly?

wkeese:
Thanks for checking.  I didn't notice anything in the reference guide (the docs/ directory).   The API guide mentioned that you couldn't change the baseClass, but I removed that text in the original commit.

wkeese:
Pushed as 8b41c394cf98dd3b65e2107dd390d1c09034b5de.

PR #523, opened 2015-02-20
Updated appium version to 1.3.4 and iOS version to 8.1 to get the iOS te...

edchat:
...sts running again.  I also had to skip 4 tests in StarRating and 1 in Switch.


wkeese:
Looks like you already pushed this as daab7113d3a4462b74ced8996578e5f9f48cb959, so closing the PR.

PR #526, opened 2015-03-16
Toaster - Adding pause message's timer on hover feature

EAlexRojas:


ToasterMessage.js
```diff
@@ -221,6 +221,44 @@ define(["dcl/dcl",
...
+		}
+
+		function _pointerLeaveHandler() {
+			if (hovering) {
+				hovering = false;
+				element._timer.resume();
+			}
+		}
+
+		this.isEnabled = false;
```
cjolif: Is that flag meant to be public? If not it should start with underscore.

ToasterMessage.js
```diff
@@ -221,6 +221,44 @@ define(["dcl/dcl",
 		}.bind(this));
 	};
 
+	var PauseTimerOnHover = function (element) {
+
+		var hovering = false;
+		function _pointerOverHandler() {
+			console.log("hover event");
```
cjolif: don't leave traces in final code.

EAlexRojas: Sorry, I'll fix that

ToasterMessage.js
EAlexRojas: Yes, as the feature can be enabled/disabled, I think it could be useful to the user to know that value.

tkrugg:
merged in https://github.com/ibm-js/deliteful/commit/286b89180bfa9e59efc32a63f29baeb640513f1f

PR #530, opened 2015-03-23
ViewStack: showPrevious() does a reverse transition

EAlexRojas:
By default showPrevious() do a reverse transition. 
If the reverse property is specified in the params, that value is used. If not, reverse = true.

### DIFF
--- a/ViewStack.js
+++ b/ViewStack.js
@@ -252,7 +252,7 @@ define(["dcl/dcl",
 				setVisibility(widget, true);
 				this._visibleChild = widget;
 
-				var transition  = (origin === widget) ? "none" : (event.transition || "slide");
+				var transition  = (origin === widget) ? "none" : (event.transition || this.transition);
 				var reverse = this.effectiveDir === "ltr" ? event.reverse : !event.reverse;
 				return this._doTransition(origin, widget, event, transition, reverse);
 			},
--- a/ViewStack.js
+++ b/ViewStack.js
@@ -169,16 +169,19 @@ define(["dcl/dcl",
 
 			/**
 			 * Shows the immediately preceding sibling of the ViewStack visible element.
-			 * The parameter 'params' is optional. If not specified, this.transition, and this.reverse are used.
+			 * The parameter 'params' is optional. If not specified, this.transition, and reverse = true are used.
 			 * @param {Object} [params] - Optional params. A hash like {transition: "reveal", reverse: true}.
 			 * The transition value can be "slide", "overlay", "fade" or "flip". Reverse transition applies to "slide"
 			 * and "reveal". Transition is internally set to "none" if the ViewStack is not visible.
+			 * Reverse is set to true if not specified.
 			 * @returns {Promise} A promise that will be resolved when the display and transition effect will have
 			 * been performed.
 			 */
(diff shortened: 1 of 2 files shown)


wkeese:
Looks fine... it's a bit rude to modify the params Object that the user passed in, but I don't know if it will matter in practice or not.

Seems like you should add a test case though.

EAlexRojas:
I did some refactoring to avoid modify the params object passed by the user in [`42d0061`](https://github.com/EAlexRojas/deliteful/commit/42d006166d728df3aa76859f4d285619dcfff2fe). What do you think?
And I added some tests in [`6390c93`](https://github.com/EAlexRojas/deliteful/commit/6390c93ff0897579fe4c7d160096939db946a020).

@cjolif I'll take it into account.

tkrugg:
I checked this PR  with the sample, and I noticed an unintuitive behaviour.
On the following screenshot, the clicking Previous (resp. Next) calls `showPrevious({})` (resp. `showNext({})`)



You can see on the screenshot that the animation is reversed. You expect the previous panel to be on the left but it slides from the right.

When i use the swiping the animation looks natural. However the panels looks like their order has been reversed. 



Here is what the DOM looks like

```html
<d-swap-view  selectedChildId="bbb">
	<div style="background-color: honeydew; " id="aaa">
		<h1>View 1</h1>
	</div>
	<div style="background-color: whitesmoke;" id="bbb">
		<h1>View 2</h1>
	</div>
	<div style="background-color: powderblue;" id=ccc>
		<h1>View 3</h1>
	</div>
	<div style="background-color: oldlace;" id="ddd">
		<h1>View 4</h1>
	</div>
</d-swap-view>
```

EAlexRojas:
I am not able to reproduce this. 
This is what I get running a sample as yours (with Previous button calling `showPrevious({})` and Next button calling `showNext({})`) :
swapview

And when swiping:
swapview2

tkrugg:
I fetched a clean version of the dependencies and am no longer experiencing the bug i reported. Sorry for the false alert.

I have one last comment: if you look at the way `showPrevious` has been implemented, both `showPrevious()` and `showPrevious({reverse: true})` will produce the same result. I find that this is inconsistent with the fact that  `showNext()` and `showNext({reverse: true})` will produce opposite results, at least from a user perspective.

```js
showPrevious: function (params) {
	// Shows the previous child in the container. 
	var args = {reverse: true};
	dcl.mix(args, params || {});
	return this._showPreviousNext("previousElementSibling", args);
}
```

Should we do something about it?

cjolif:
> Should we do something about it?

Probably?

dmandrioli:
>Should we do something about it?

I think it is good enough as-is. These 2 methods are shortcuts for the show() method in the common use-case of navigating gradually in a set of views.
Despite it is possible to pass {reverse:XXX} to showPrevious/showNext, it does not look to be a common use-case. I guess showNext and showPrevious are used 99% of the time without argument or with an argument on the transition type (fade, slide, etc).

cjolif:
ok, so we should merge this? :)

dmandrioli:
>ok, so we should merge this? :)

Yes

Issue #541, opened 2015-04-02
remove calls to parse() and attachedCallback() from samples, tests


wkeese:
The sample files no longer need to call `parse()`.  As a corollary, most of them don't need to `require()` `"delite/register"` explicitly anymore either.

The functional tests should call `register.deliver()` rather than `register.parse()`.   The `deliver()` call is only necessary so that the `ready` flag doesn't get set prematurely.

EAlexRojas:
When testing this changes I found a problem using the latest version of dcl (1.1.3).
On IE11 none of the samples/test works due to a change introduced in that dcl version. 
I created an issue against dcl to report the problem: 
[https://github.com/uhop/dcl/issues/18 ](https://github.com/uhop/dcl/issues/18)

AdrianVasiliu:
In the Combobox unit test, 9e32a0e8f2e13c765252268d9bd8241d093558e3 replaced `register.parse(container);` by `register.deliver();`. Apparently, this change has something to do with the new errors that appeared for IE10 in the travis build (there used to be no error in the unit test):

```
STARTED  Test 'main - deliteful/Combobox: markup - Store.add/remove/put (custom element store)' 
on internet explorer 10 on WINDOWS:
undefined: Unable to get property 'nodeType' of undefined or null reference
TypeError: Unable to get property 'nodeType' of undefined or null reference
  at getEnclosingWidget  <delite/Widget.js:296:5>
  at getParent  <delite/Widget.js:226:4>
  at attachedCallback  <deliteful/list/ItemRenderer.js:25:17>
  at Anonymous function  <delite/register.js:372:16>
  at f  <dcl/advise.js:72:36>
  at upgrade  <delite/register.js:137:6>
  at Anonymous function  <delite/register.js:499:7>
  at processMutations  <delite/register.js:493:4>
  at deliver  <delite/register.js:536:5>
  at beforeEach  <deliteful/tests/unit/Combobox.js:281:4>
```
(https://travis-ci.org/ibm-js/deliteful/builds/59724806).

The direct cause of the error is that Widget.getParent() now gets called for a widget which, on IE10, is not (yet?) in the DOM. Thus, the implementation of Widget.getParent() (`return this.getEnclosingWidget(this.parentNode);`) calls Widget.getEnclosingWidget with a null argument, which is not expected by getEnclosingWidget()... 
[...]

PR #542, opened 2015-04-09
modif samples : remove register.parse

harbalk:
I made the corrections on a part of the samples as asked in this issue : https://github.com/ibm-js/deliteful/issues/541
I have tested it on Chrome, Firefox, IE11, Safari (on iOS8) and Android (4.X).
I didn't find any regression.
There is however a problem on Android devices with Android browser where any tests is running (but works perfectly on Chrome for Android and Firefox).
Also one little problem on StarRating.html where green-man style doesn't work on Chrome for Android.

### DIFF
--- a/samples/Combobox.html
+++ b/samples/Combobox.html
@@ -19,7 +19,6 @@
 
 	<script type="text/javascript">
 		require([
-			"delite/register",
 			"dstore/Memory",
 			"deliteful/list/List",
 			"deliteful/Combobox",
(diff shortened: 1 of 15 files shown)


edchat:
Which Android devices, and versions were having problems?

harbalk:
On HTC One M8 4.4 and Samsung Galaxy S4 4.2.2 "Internet" Browser. (Like I said it works with Chrome and Firefox).
But the tests weren't running before the modifications of parse call so I don't think this come from that..

cjolif:
> But the tests weren't running before the modifications of parse call so I don't think this come from that..

When I was asking for possible regressions I was meaning against delite 0.6. Not just after the parse() call removal.

wkeese:
Also, you mentioned that nothing works on IE11 with the latest dcl.  Did you file a ticket for that?

harbalk:
It was EAlexRojas who had the problem with the latest dcl. I don't have the informations.

wkeese:
Oh OK.  The changes to the samples look good.  I assume someone is doing the tests separately.

edchat:
Merged with https://github.com/ibm-js/deliteful/commit/4c678a501b1f85f489d4bc33b51a5beee6dde1b3

PR #544, opened 2015-04-10
Removing calls to register.parse from samples and using deliver on functional tests

EAlexRojas:
Adaptation of some samples and tests to remove the register.parse calls and to use deliver instead of register on functional tests, as indicated in  [#541](https://github.com/ibm-js/deliteful/issues/541).

Tested on latest Chrome, Firefox and IE11 on desktop. Safari on iOS 8.1, Chrome, Firefox and default ("Internet") browser on Android 4.2.2 (Samsung Galaxy S4) and Chrome on Android 4.4 (Nexus  5).

On desktop: no regressions found except for a problem in a specific sample: "progressBar-handler.html". 
When the register.parse() call is removed, the error "progressBar.observe is not a function" is thrown one in two times. (Refers to [this line](https://github.com/ibm-js/deliteful/blob/f6b78d7be72068140e8c2d957ffc2e5ac4c080f1/samples/ProgressBar-handler.html#L62) )

Also, none of the samples runs on IE11. Apparently the problem is due to a change in the latest version of dcl (1.1.3) (Refers to [this commit](https://github.com/uhop/dcl/commit/b4b492c2f59aaef2dfec5830107cf0761fc484c1)). 
At that point f is the HTMLElement to register and that typeof HTMLElement usually returns "function", but in IE11 it returns "object". 

On iOs: no regression found.

On Android: 
4.4: works perfectly on Chrome.
4.2: works perfectly on Chrome and Firefox. but none of the samples works on the default ("Internet") browser. (even without removing the register.parse() call)

### DIFF
--- a/samples/Button.html
+++ b/samples/Button.html
@@ -18,12 +18,10 @@
 
 	<script type="text/javascript">
 		require([
-			"delite/register",
 			"deliteful/Button",
 			"delite/theme!delite/themes/{{theme}}/global.css",	// page level CSS
 			"requirejs-domready/domReady!"
-		], function (register) {
-			register.parse();
+		], function () {
 			document.body.style.display = "";
 		});
 	</script>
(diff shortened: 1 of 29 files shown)


cjolif:
> On desktop: no regressions found except for a problem in a specific sample: "progressBar-handler.html". 
When the register.parse() call is removed, the error "progressBar.observe is not a function" is thrown one in two times. (Refers to this line )

Sorry if that was not clear but I was also wanted you to check regressions against 0.6 not just after the parser call removal.

edchat:
Merged with https://github.com/ibm-js/deliteful/commit/19399079349dd1257fde61d9b70d4721d302ecac

PR #547, opened 2015-04-14
Use dcl 1.1.2 temporarily to get around problem with 1.1.3, update inter...

edchat:
...n to 2.2.2, and skip 2 Combobox tests on IE10 for now.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
 	"name": "deliteful",
 	"version": "0.7.0",
 	"dependencies": {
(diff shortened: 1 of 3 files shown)


cjolif:
This change should be done accordingly for all ibm-js projects we will re-issue this release to avoid conflicts (if there are other projects with a dcl dependency which I suspect ;)).

edchat:
Ok, this change did not cause a conflict because the others are using 1.1.x, so it used 1.1.2.

PR #550, opened 2015-04-16
Modification of call parse to call deliver on functionnal tests

harbalk:
All the tests are passing on Chrome, Firefox and IE11.


wkeese:
Then end result (https://github.com/ibm-js/deliteful/pull/550/files) seems fine.  I don't know why this PR has 5 commits with strange changes, but I guess those will go away when its squashed into a single commit and pushed to ibm-js.

harbalk:
Yes I'm sorry I pushed modifications I didn't want to. So I have have to remove them with other commits.

edchat:
I pushed this with https://github.com/ibm-js/deliteful/commit/772541f3dbb62f90b3f864090185e61177881749  
When I tried to set --author="harbalk" it did not take it, so I was not able to set the author.

PR #552, opened 2015-04-20
remove call parse to deliver in unit tests

harbalk:
Modifications of the unit tests to call deliver instead of parse.

### DIFF
--- a/tests/test_Button.html
+++ b/tests/test_Button.html
@@ -25,7 +25,7 @@
 				}
 			});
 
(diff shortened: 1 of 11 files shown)


edchat:
register.deliver() does not take any params, so I will remove those before I push this.

edchat:
Pushed with https://github.com/ibm-js/deliteful/commit/5ecdf3bf4b5118864f340738b8fdf118470171c7

PR #553, opened 2015-04-20
Using register.deliver() rather than register.parse() on unit tests

EAlexRojas:


edchat:
Pushed with https://github.com/ibm-js/deliteful/commit/9e32a0e8f2e13c765252268d9bd8241d093558e3

Issue #554, opened 2015-04-21
Some widgets don't fully work with touch gesture on IE11 Mobile, WP 8.1


EAlexRojas:
When the screen is pressed and the pointer is held on the screen, pointerUp and touchEnd events are fired (while still holding the pointer down). 
After that, neither move events nor the real pointerUp event can be captured while screen is held pressed. 
(That can be tested using dpointer/tools/tracker/tracker.html)

This affects the normal behaviour of some widgets like toasterMessage, swapView, slider, startRating and scrollableContainer.

This bug is already reported to microsoft here: https://connect.microsoft.com/IE/feedback/details/1076515/mouseup-pointerup-events-are-fired-automatically-when-mouse-or-pointerdown-are-held-windows-phone-8-1-ie11

Device used to test: Nokia Lumia 920 with Windows Phone 8.1u1 (8.10.14234.375) and IE11.

PR #555, opened 2015-04-22
ToasterMessage: Timer is resumed on pointerup on touch devices

EAlexRojas:
For input devices that do not support hover, a pointerover event is fired before the pointerdown event, to avoid message who stay forever on the screen, the timer is resumed on pointerup when the pointer type is touch.

Note: on Window Phone 8.1 with IE11, when pointer is held pressed on the message, a pointerup is fired and the timer is resumed, due to the problem referenced on https://github.com/ibm-js/deliteful/issues/554

I changed messages placement to bottom rigth on toaster sample to have a better presentation on mobile devices.

### DIFF
--- a/ToasterMessage.js
+++ b/ToasterMessage.js
@@ -224,17 +224,23 @@ define(["dcl/dcl",
 	var PauseTimerOnHover = function (element) {
 
 		var hovering = false;
-		function _pointerOverHandler() {
+		function _pauseTimer() {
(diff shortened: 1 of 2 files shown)


tkrugg:
merged in https://github.com/ibm-js/deliteful/commit/681a32a2a5da46cdce6ef40c7d33d470ad040487

PR #557, opened 2015-04-22
Adding register.deliver() call in samples that need it. Fixes #556

### LINKED ISSUE #556: Some samples need to call register.deliver()
In some samples we need to be sure that the element is  already a widget in order to use methods like "observe" and "on" or to get access to its members, so a call to `register.deliver()` is needed. 
The samples are:

* [ProgressBar-handler](https://github.com/ibm-js/deliteful/blob/master/samples/ProgressBar-handler.html#L62)
* [Slider](https://github.com/ibm-js/deliteful/blob/master/samples/Slider.html#L31)
* [ViewStack](https://github.com/ibm-js/deliteful/blob/master/samples/ViewStack.html#L35)
* [SwapView](https://github.com/ibm-js/deliteful/blob/master/samples/SwapView.html#L28)

(At least on Firefox those samples works only one in two times)

EAlexRojas:


wkeese:
Maybe it's better to use the declarative syntax for specifying event handlers:

```xml
<my-widget on-click="console.log(event);"
	on-element-selected="console.log(event);"></my-widget>
```

Then you don't need the `register.deliver()` call except for one file (apparently)

EAlexRojas:
The `register.deliver()`call is needed in ProgressBar-handler to use the `observe()` method and in the Slider sample to get acces to `slider.progressBar` and to use the `observe()` method.

I used the declarative syntax for specifying the event handlers on c20bae2613393c3f3ebb0a82d30499b04ef70364

However, this works because the function used in the handlers is console.log. 
If the user want to use its own function, the function should be created before the widget initialization... that is before the require call (?)

Note: ViewStack doesn't use hide() method, so I think the "delite-before-hide" and "delite-after-hide" handlers are not needed.

wkeese:
OK.

> If the user want to use its own function, the function should be created before the widget initialization... that is before the require call (?)

Hmm, for  `require([...], callback)`, it's true that the widget initialization could happen either before `callback()` is executed or not until `callback()` calls `register.deliver()`.   But you only need to make sure the function is defined before the event happens.  In other words, `on-click="foo()"` presumably won't execute before `callback()` completes.  Hope that makes sense.

EAlexRojas:
Ok, that makes sense. But in the case of events like `delite-display-load` fired by ViewStack, the event could be fired before `callback()` (?)

wkeese:
Yes, I think so.  Of course, that's true regardless of how you set up the event listener.

edchat:
Pushed as https://github.com/ibm-js/deliteful/commit/a62d11aa1fcb47dbb15f441afdfb25c21968a784
Thanks @EAlexRojas

PR #559, opened 2015-04-29
Updating ScrollableContainer unit test to use register.deliver() instead of register.parse()

EAlexRojas:


edchat:
Push with https://github.com/ibm-js/deliteful/commit/2d9122a84f27539cf44558917405bccc67625fde
Thanks

Issue #551, opened 2015-04-16
ProgressBar: message appears twice when progressBar has an initial value


EAlexRojas:
When a ProgressBar is created with an initial value the message appears in two different places:

![progressbar](https://cloud.githubusercontent.com/assets/2373303/7177449/0e75f27e-e427-11e4-8f7b-15cb318e3802.PNG)

The width of msgInvertNode (white text) is changed in refeshRendering method in order to have the same width than msgNode using getComputedStyle.  ([here](https://github.com/ibm-js/deliteful/blob/master/ProgressBar.js#L134))

This works on 0.6.0.

I found that on 0.6.0 version, refreshRendering is called when there's already something displayed on the screen, but on actual version, when refreshRendering is called there's nothing  displayed yet on the screen. (so, getComputedStyle doesn't returns the desired value)

Not sure if it's a delite problem or it's suposed to be like that and so it's a progressBar problem. 

@wkeese what do you think about it? 

Code used: 
```javascript
<d-progress-bar value=40 max=100></d-progress-bar>
````

**Comment by wkeese** [comment]
This needs to be fixed in ProgressBar.  refreshRendering() shouldn't assume that the widget is attached to the document.  That's what attachedCallback() is for.

I guess you are talking about this code:

```js
if ("value" in props || "max" in props) {
	if (this.position === -1) { //indeterminate state
		this.indicatorNode.style.removeProperty("width");
		this.removeAttribute("aria-valuenow");
	} else {
		this.indicatorNode.style.width = (this.position * 100) + "%";
		this.msgInvertNode.style.width =
			window.getComputedStyle(this.msgNode).getPropertyValue("width");
		this.setAttribute("aria-valuenow", this.value);
	}
}
```

I don't understand why it's calling `this.indicatorNode.style.removeProperty("width");` when `position === -1`, but if that's really needed, it makes things more complicated.  Perhaps attachedCallback() can set the width, and refreshRendering() can toggle `style.display` between `block` and `none`.

wkeese:
PS: I got confused in my last comment between `indicatorNode` and `msgInvertNode`.  It seems like all you need to do to fix this problem is to move:

```js
        this.msgInvertNode.style.width =
            window.getComputedStyle(this.msgNode).getPropertyValue("width");
```

to attachedCallback()?

EAlexRojas:
Even doing that on attachedCallback I have the same result.

Based on the samples, I'm using this code to test:
 ```javascript
...
[...]

PR #561, opened 2015-05-06
ProgressBar: Changes on sample to work around #551

EAlexRojas:
This is a temporary change to work around the issue #551


wkeese:
Seems fine to me.

Issue #569, opened 2015-05-19
Problem on PageableList when modifying items at index to high


harbalk:
I noticed a problem when doing an update on an item of index greater than pageLength * maxPages.

It is because the function getItemRenderers() of List only returns the items displayed. Also it returns maximum pageLength * maxPages items. Then the expression 'this.getItemRenderers().item(index)' return null if the index is too high.

Moreover if we modified the item 0 but the first item displayed is the 5th then it will be the 5th that will be modified at first time on the display. (When we reload the page it is allright because the data is not modified, only the display).

As you can see it is just a question of shifting in the indexes, a solution is to overwrite the function getItemRendererByIndex of List in PageableList to treat the index correctly.

What do you think about it ?

PR #570, opened 2015-05-20
modifications on getItemRendererByIndex for PageableList

harbalk:
Proposal to fix https://github.com/ibm-js/deliteful/issues/569

### DIFF
--- a/list/PageableList.js
+++ b/list/PageableList.js
@@ -447,9 +447,9 @@ define([
 			var idPage, i;
 			if (first) {
 				idPage = this._idPages.shift();
-				this._firstLoaded += idPage.length;
 				for (i = 0; i < idPage.length; i++) {
-					this._removeRenderer(this.getItemRendererByIndex(0), true);
+					this._removeRenderer(this.getItemRendererByIndex(this._firstLoaded), true);
+					this._firstLoaded++;
 				}
 				if (idPage.length && !this._previousPageLoader) {
 					this._createPreviousPageLoader();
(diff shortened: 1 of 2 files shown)


wkeese:
Shouldn't this have a test?   Or are the existing tests already failing?

harbalk:
There was a test but it didn't use the function correctly. (It is the tests modified in the commit).
The test was asking the item at index 0 to get the first item displayed (item 10 in the test) but not the item which is at the first index in the list (item 0).

harbalk:
Reading the description of the function again, I think the question is to know what do we mean by index.
If it is the index in the item displayed (equivalent to the itemRenderers) the old version of the function was correct.
But we shouldn't use it in the function itemRemoved and itemUpdated in List.js, where the index used is the index of the item in the list.

cjolif:
@harbalk is your PR on ObservableArray depending on this one? Or can it work even without this fix?

harbalk:
@cjolif  it can work without this fix.

cjolif:
was that closed by mistake or has the fix been merged? (if yes a link to the commit ID would be nice).

harbalk:
The fix has not been merged. I dig into it again and I realize the problem occured only when it is an itemUpdated. Also the problem was not coming from getItemRendererByIndex but from PageableList which overwrite itemAdded and ItemRemoved but not itemUpdated.
I made the correction on another branch I will do the PR in a few minutes.

cjolif:
ok, next time leave comments when closing so that I don't wonder what's happening ;) thanks.

PR #571, opened 2015-05-20
Accordion widget

EAlexRojas:
This is **not fully ready yet**, it's only here to gather some feedback on the work made until now.

* I created a Panel Container in order to have a container with a title bar and the Accordion only accepts 
Panels as children.

* The Accordion has a fixedHeight (100% of the parent by default). Only one or multiple panels can be open at
a time, but at least one remains always open to fill the available space. (I'm using CSS3 Flexible box to achieve that)

* Animations on the flex property are not supported on IE 11 and before.

What remains to be done:

* Adding unit and functional tests
* Moving between panels using arrow keys
* Adding support for screen reader
* Adding holodark and ios themes for the widgets ?
* Animationend event detection refactoring


wkeese:
With dijit's accordion we struggled with whether the "Panel Container" should be part of the API, or just implemented as something internal.   We eventually decided to make it internal.   And @cjolif also took the same approach with his Bar.js class, as seen in his in-progress work on https://github.com/cjolif/deliteful/commit/5ced15d0fba240a530d186a2cc76cfa22dad3c86#commitcomment-10802199.   Notice how he overrides addChild() etc.

So, probably your work should be combined with his, and the Panel Container should be created/removed behind the scenes?  His Bar.js base class could be something more generic (and perhaps in delite).   Anyway, something to think about.

Other than that your description sounds reasonable to me.   I'll take a look at the code though and add line-item comments if I see something.

EAlexRojas:
> With dijit's accordion we struggled with whether the "Panel Container" should be part of the API, or just implemented as something internal. We eventually decided to make it internal. And @cjolif also took the same approach with his Bar.js class ...
> So, probably your work should be combined with his, and the Panel Container should be created/removed behind the scenes?

I thought about that when I started the development of the accordion, and I decided to create the panelContainer in order to have control of the titleNode, the contentNode and the properties like the title of the panel and the icons to use when the panel is closed or open.

If a widget is not used to represent the panel, we will have as children of the accordion: arbitrary `div` with those properties, or widgets with those additional properties. 
I know that could be possible, but I don't know if that's what we want. ?

In the case we decide to use this approach, this accordion could be similar to the Bar.js of @cjolif, but here, in addition to wrap each child, it's necessary to add a toggle for each of them.

I think is better to have a widget to manage the relationship an reference between the toggle and the content of each panel or what do you think about it ?

wkeese:
> If a widget is not used to represent the panel, we will have as children of the accordion: arbitrary div with those properties, or widgets with those additional properties. 
I know that could be possible, but I don't know if that's what we want. ?

Yeah, I don't know either.  It's definitely more complicated for you.  On the flip side, it's sometimes simpler for the application developer, since he does:

```xml
<d-accordion>
    <d-scrollable-container label="pane 1">
          ...
   </d-scrollable-container>
</d-accordion>
```

rather than:

```xml
<d-accordion>
    <d-panel label="pane 1">
       <d-scrollable-container>
          ...
       </d-scrollable-container>
   </d-panel>
</d-accordion>
```

Programmatic construction also becomes less tedious.  But, at least your approach is KISS.

@edchat - Would having that intermediate `<d-panel>` make things too complicated for the dapp code?

BTW, `<d-panel>` seems like too generic a name, because don't lots of widgets have panels, for example ViewStack?

edchat:
@wkeese good question, I think dapp would be ok, but we might need a way to
set (or figure out) the id (or at least a selector) for the inner <d-panel>
if that is where the dapp view was to be placed.  The id is probably not
required, but we need to be able to setup a parentSelector for the view to
be able to determine the parent node.

On Thu, May 28, 2015 at 7:38 PM, Bill Keese <notifications@github.com>
wrote:

> If a widget is not used to represent the panel, we will have as children
> of the accordion: arbitrary div with those properties, or widgets with
> those additional properties.
> I know that could be possible, but I don't know if that's what we want. ?
>
> Yeah, I don't know either. It's definitely more complicated for you. On
> the flip side, it's sometimes simpler for the application developer, since
> he does:
>
> <d-accordion>
>     <d-scrollable-container label="pane 1">
>           ...
>    </d-scrollable-container>
> </d-accordion>
>
> rather than:
>
> <d-accordion>
>     <d-panel label="pane 1">
>        <d-scrollable-container>
>           ...
>        </d-scrollable-container>
>    </d-panel>
> </d-accordion>
>
> Programmatic construction also becomes less tedious. But, at least your
> approach is KISS.
>
> @edchat <https://github.com/edchat> - Would having that intermediate
> <d-panel> make things too complicated for the dapp code?
>
> BTW, <d-panel> seems like too generic a name, because don't lots of
> widgets have panels, for example ViewStack?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ibm-js/deliteful/pull/571#issuecomment-106629757>.
>



-- 
Ed Chatelain

wkeese:
OK... and possibly also code for creating a `<d-panel>` to hold a dapp view?

EAlexRojas:
>BTW, `<d-panel>` seems like too generic a name, because don't lots of widgets have panels, for example ViewStack?

We could maybe use something like `<d-titled-panel>` ?

wkeese:
Well, that sounds like something that could be used outside of Accordion though, like http://download.dojotoolkit.org/release-1.10.2/dojo-release-1.10.2/dijit/tests/test_TitlePane.html.   `<d-accordion-panel>` seems better to me.  Just my opinion.

cjolif:
@wkeese I guess that's the whole point, I think @EAlexRojas panel can be used outside? Like if you want a titled panel in other contexts?

Accordion.js
```diff
@@ -0,0 +1,474 @@
...
+		 */
+		selectedChildId: "",
+
+		/**
+		 * If true, only one panel is open at a time.
+		 * If false, several panels can be open at a time, but there's always at least one open.
+		 * @member {boolean}
+		 * @default true
+		 */
+		singleOpen: true,
```
cjolif: It might be more future proof to not use a boolean value. For example if one day we want to include a 3rd mode (like ability to close all panels) we will be stuck. For this reason, except in cases where I don't foresee a possible 3rd mode for the future, I usually tend to prefer a "mode" with various values. That said I don't have a very strong opinion.

Accordion.js
```diff
@@ -0,0 +1,474 @@
...
+		 */
+		closedIconClass: "",
+
+		_panelList: [],
+
+		_numOpenPanels: 0,
+
+		_changeHandler: function (event) {
+			var panel = event.target.parentNode;
+			//Case when the event is fired by the label or the icon
```
cjolif: I think we tend to have spaces between // and comments. Might be good for consistency to do the same.

Accordion.js
```diff
@@ -0,0 +1,474 @@
...
+
+		/**
+		 * The default CSS class to apply to DOMNode in children headers to make them display an icon when they are
+		 * closed. If a child panel has its own closedIconClass specified, that value is used on that panel.
+		 * @member {string}
+		 * @default ""
+		 */
+		closedIconClass: "",
+
+		_panelList: [],
```
cjolif: We should avoid assigning objects to fields to avoid the risk of it being shared by all instances as part of the prototype.

Accordion.js
```diff
@@ -0,0 +1,474 @@
...
+						if (panel.open && panel !== this._selectedChild) {
+							this.hide(panel);
+						}
+					}.bind(this));
+				}
+			}
+		},
+		/* jshint maxcomplexity: 10 */
+
+		_supportAnimation: function () {
```
cjolif: isn't that naming a bit misleading? I mean it is also checking if we are in a condition where will not work (invisible) even if animation is actually supported? (that's detail).

Accordion.js
EAlexRojas: OK. I'll fix that.

Accordion.js
EAlexRojas: Maybe we could change the name of the method to `_useAnimation()` and move into it also the validation of `this.animate` ?

Accordion.js
EAlexRojas: OK. I'll change that in all the comments.

Accordion.js
EAlexRojas: I agree, it would be better to use a `mode` property and give the possibility to eventually add more modes in the future.

EAlexRojas:
PR updated taking @cjolif comments into account

Accordion.js
```diff
@@ -0,0 +1,496 @@
...
+	"delite/DisplayContainer",
+	"./ToggleButton",
+	"delite/theme!./Accordion/themes/{{theme}}/Accordion.css"
+], function (dcl, has, Promise, register, KeyNav, $, DisplayContainer, ToggleButton) {
+
+	function setVisibility(node, val) {
+		node.style.display = val ? "" : "none";
+	}
+
+	var animationEndEvent = (function () {
```
tkrugg: I don't remember if we've already discussed this but this 'hack' is becoming some sort of pattern which many animated widgets are likely to need. We might want to consider putting it somewhere it can be shared.

Accordion.js
```diff
@@ -0,0 +1,496 @@
...
+	/* Accordion modes */
+	var accordionModes = {
+		singleOpen: "singleOpen", // default
+		multipleOpen: "multipleOpen"
+	};
+
+	var defaultMode = accordionModes.singleOpen;
+
+	/**
+	 * A layout container that display a vertically stacked list of Panels whose titles are all visible, but only one
```
tkrugg: display**s**

Accordion.js
```diff
@@ -0,0 +1,496 @@
...
+		 * The id of the panel to be open at initialization.
+		 * If not specified, the default open panel is the first one.
+		 * @member {string}
+		 * @default ""
+		 */
+		selectedChildId: "",
+
+		/**
+		 * The mode of the Accordion
+		 * `mode` is one of `["mode", "multipleOpen"]`.
```
tkrugg: one of ["singleOpen", ...]

Accordion/themes/Accordion_template.less
```diff
@@ -0,0 +1,106 @@
+@import "../../../../delite/themes/bootstrap/includes";
+
+.d-accordion {
```
tkrugg: detail: it should be tab indented I guess, although we do have .less files that aren't elsewhere...

docs/Accordion.md
```diff
@@ -0,0 +1,255 @@
+---
+layout: doc
+title: deliteful/Accordion
+---
+
+# deliteful/Accordion
+
+``deliteful/Accordion`` is a layout container that display a vertically stacked list of Panels whose titles are all
```
tkrugg: display**s**

docs/Accordion.md
```diff
@@ -0,0 +1,255 @@
...
+of the form `{child: HTMLElement}`.
+The event also provides the method `setContent()` that must be used to specify the content for the panel.
+It's important to note that `evt.dest` make reference to the Panel id and `evt.contentId` to the content id.
+
+The user would have access to this through the `show()` and `hide()` methods. e.g; `show("p1", {contentId: newContent})`.
+The example could be changed to load new content even if the panel already exists. See samples/Accordion-controller.
+
+
+Another interesting behaviour that can be achieved by using a controller is creating empty panels and loading their content
+only when the panel be opened. In the following example there's a way to achieve that:
```
tkrugg: is opened

docs/Accordion.md
```diff
@@ -0,0 +1,255 @@
...
+
+<a name="enterprise"></a>
+## Enterprise Use
+
+### Accessibility
+
+|type|status|comment|
+|----|------|-------|
+|Keyboard|ok|It is keyboard navigable (see [User Interactions](#interactions))|
+|Visual Formatting|ok|Support high contrast on Firefox and Internet Explorer desktop browsers.|
```
tkrugg: Support**s** I guess, since you're using *It* in the row before?

docs/Panel.md
```diff
@@ -0,0 +1,115 @@
...
+
+<a name="enterprise"></a>
+## Enterprise Use
+
+### Accessibility
+
+|type|status|comment|
+|----|------|-------|
+|Keyboard|N/A|No user interaction|
+|Visual Formatting|ok|Support high contrast on Firefox and Internet Explorer desktop browsers.|
```
tkrugg: same note on "support"

Accordion.js
EAlexRojas: Maybe we should add a feature test for animation in order to be able to do `has("animation")` or something like that ?
Some widgets like Viewstack and SidePane only test if the browser is not IE9 to validate if it's possible to perform transitions.
What do you think @wkeese ?

wkeese: I agree... well this code is already doing feature testing, it's just that it's got a strange format of

```
var animationEndEvent = (function () { ... }());
```

instead of

```
has.add("animationEndEvent", function () { ... });
```

EAlexRojas: Ok, but maybe we should put that feature test somewhere it can be used by all the widgets that need it ?

cjolif: Yes, let's move it to features.js and use it everywhere it is used.

cjolif:
merged in 46eb34099b7bf86ab140e606c4978879c0f514ec

PR #573, opened 2015-05-22
Possibility to use ObservableArray in Store instead of a dstore/Store (first draft) deliteful

harbalk:
**This pull request is just made to have some feedback on the work I have done until now**. It is not complete for the moment.

The goal is to approach the current use of delite/Storewith an array instead of a dstore/Store : https://github.com/ibm-js/delite/issues/278

What have been done : 
- Documentation to describe the use of lists with an array in store property
- Unit tests on List and PageableList with arrays
- Few modifications on PageableList.js and List.js 

What still remaining : 
- Replace the delitful/Store to create a store with an array instead of `dstore/Memory`.
- Communication between the array and a server


wkeese:
I looked over this briefly.   Seems good; I don't have any other comments besides what I wrote in delite.

tests/unit/list/resources/ListBaseTestsArrayObs.js
```diff
@@ -0,0 +1,407 @@
...
+						var focusedElement = document.activeElement;
+						assert.isNotNull(focusedElement, "active element");
+						assert.isDefined(focusedElement, "active element");
+						assert.strictEqual("item 1", focusedElement.parentNode.item.label, "focused element label");
+						d.resolve();
+					}));
+					return d;
+				},
+				teardown : function () {
+					//this.list.destroy();
```
cjolif: If not needed, I guess we can just delete those comments.

harbalk:
Update the PR with the new adapters and renaming of store property : https://github.com/harbalk/deliteful/commit/2241f1ecb526ec0ae333f235681a1d9103e5c997

Combobox.js
```diff
@@ -482,8 +480,9 @@ define([
 				
 				if (!initValueSingleMode()) {
 					// List not ready, wait.
-					this.list.on("query-success", function () {
+					var waitListener = this.list.on("query-success", function () {
```
cjolif: Why do we need that waitListener that we did not need before?

Select.js
```diff
@@ -239,7 +238,9 @@ define([
 						option.__dataItem = renderItem.__item; // __item is set by StoreMap.itemToRenderItem()
 						// to allow retrieving the option element from widget's selectedItems
 						// (which are data items, not render items).
-						option.__dataItem.__visualItem = option;
+                        if (!option.__dataItem.__visualItem) {
```
cjolif: Looks like there might be some alignement issue here.

Select.js
```diff
@@ -290,10 +291,11 @@ define([
 			}
 		},
 		
-		getIdentity: function (dataItem) {
-			// Override of delite/Selection's method
-			return this.store.getIdentity(dataItem);
-		},
+		getIdentity: dcl.superCall(function (sup) {
+            return function (dataItem) {
+                return sup.call(this, dataItem);
```
cjolif: same tabbing/alignement issue. Please check it everywhere (isn't jshint raising errors?)

Combobox.js
harbalk: The variable waitListener is needed only to be able to remove the listener after the first initialisation.
It was not needed before because the firstItemRenderer was already loaded when the function was called. Also the ```initValueSingleMode()``` function was always returning true and the listener was not defined.
The problem was when modifying the query (with autoFilter === true), when supressing a character the function ```initValueListener()``` was automatically called and the label of the first item was refreshed. Then we could not see the query anymore.

Select.js
harbalk: The indentation is fine in my file so jshint isn't raising errors.
But I think it is a problem with my tabs which are not well interprated by github. I'll fix it.

Select.js
```diff
@@ -65,7 +64,7 @@ define([
 	 * @augments module:delite/StoreMap
 	 * @augments module:delite/Selection
 	 */
-	return register("d-select", [HTMLElement, FormWidget, StoreMap, Selection],
+	return register("d-select", [HTMLElement, FormWidget, Selection, StoreMap],
```
clmath: If the mixin order is important, it may worth adding a comment.

Select.js
```diff
@@ -239,7 +238,9 @@ define([
 						option.__dataItem = renderItem.__item; // __item is set by StoreMap.itemToRenderItem()
 						// to allow retrieving the option element from widget's selectedItems
 						// (which are data items, not render items).
-						option.__dataItem.__visualItem = option;
+						if (!option.__dataItem.__visualItem) {
```
clmath: Why is it needed to change the priority between option.__dataItem.__visualItem and option ? Shouldn't option.__dataItem = renderItem.__item be changed too ?

samples/list/List-ObservableRest.html
```diff
@@ -0,0 +1,115 @@
...
+			'requirejs-dplugins/Promise!', // page level CSS
+			"deliteful/list/List",
+			"decor/ObservableArray",
+			"decor/Observable"
+		], function (request, register, Promise, List, ObservableArray, Observable) {
+
+			var list  = register.createElement("d-list");
+			list.id = "L";
+
+			var initRequest = request("http://9.128.82.26:3000/book", {
```
clmath: Using a variable to store the rest endpoint would allow a user to customize it more easily.

samples/list/PageableList-ObservableRest.html
```diff
@@ -0,0 +1,95 @@
...
+			"delite/register",
+			'requirejs-dplugins/Promise!', // page level CSS
+			"deliteful/list/PageableList",
+			"decor/ObservableArray"
+		], function (request, register, Promise, PageableList, ObservableArray) {
+
+			var list  = register.createElement("d-pageable-list");
+			list.id = "L";
+
+			var initRequest = request("http://9.128.82.26:3000/book?&start=" + 0 + "&length=" + 4, {
```
clmath: Using a variable to store the rest endpoint would allow a user to customize it more easily.

cjolif:
Looks like all comments have been taken into account, any last minute one before I merge?

cjolif:
merged in 11a4c8d33dd91d816d4b10beef17ecc57dd0fb08

Issue #574, opened 2015-06-05
Use of removed delite/key.js file in deliteful.


harbalk:
The file delite/key.js in still used in some of deliteful files (Combobox, Slider, StarRating, SwapView and list/List).
This file has been removed in this commit https://github.com/ibm-js/delite/commit/cac5fadea9248a6e495a18b4986c9455964c6b20.
Then there are errors when using require.js.

PR #575, opened 2015-06-18
remove delite/keys.js from deliteful

harbalk:
Fix https://github.com/ibm-js/deliteful/issues/574

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -7,14 +7,12 @@ define([
 	"delite/register",
 	"delite/FormValueWidget",
 	"delite/HasDropDown",
-	"delite/keys",
 	"./list/List",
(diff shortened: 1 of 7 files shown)


wkeese:
Looks reasonable to me.  If you check this in now though you need to update the bower.json to point to delite master rather than delite 0.7.x.

harbalk:
Ok thanks, done on 597fbe3f72ab2989ed85813ed28bd47eb31b97b6

wkeese:
OK, I pushed this in abdffa9db18513e8e1e0ba16482efa8401249e91 and b17eff1cb92f7b5715400b775a661a08f658cc71.   I changed the bower.json to say "master" rather than "#master".  AFAICT the "#" is unnecessary.  But maybe I'm wrong.

PR #576, opened 2015-06-18
Keeping a clean DOM on the visible view after shown it. Fixes #567

### LINKED ISSUE #567: SwapView: After calling show method with transition, Swipe transitions are broken
After showing a view using a transition, swipe animation doesn't work correctly.
On the following screenshot, clicking Previous (resp. Next) calls ``showPrevious({})`` (resp. ``showNext({})``)

swapview - new

Before clicking the button, swipe animations works perfectly, but after clicking the button, the visible view doesn't move anymore. 

Apparently, after performing the show method, the visible view still have some CSS classes related to the transition, (in this case ``-d-view-stack-slide -d-view-stack-in -d-view-stack-transition``) and that interferes with the transitions of the swipe. (more specifically with the ``transform`` property)

EAlexRojas:
- Fixing regression produced since passing from using dojo/Deferred to Pomise
 - Updating tests according to this change

### DIFF
--- a/ViewStack.js
+++ b/ViewStack.js
@@ -233,7 +233,10 @@ define(["dcl/dcl",
 						setVisibility(origin, false);
 					}
 				}
-				return Promise.resolve(promise);
+				return Promise.resolve(promise).then(function () {
+					cleanCSS(this._visibleChild);
+					$(this).removeClass("-d-view-stack-transition");
+				}.bind(this));
 			},
 
 			changeDisplay: function (widget, event) {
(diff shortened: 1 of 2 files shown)


dmandrioli:
The original code was assuming that transitionend event could randomly be dropped for one of the two simultaneous transitions.
That was the sense of the workaround, the first transitionend triggers delite-after-show, the second triggers DOM cleanup.
In this implementation, only the second event is taken into account to trigger delite-after-show and cleanup the DOM.

If transitionend can still be dropped erratically, the regression is not fixed.

@EAlexRojas suggests that setting display:none could avoid transitionend to occur.

If it is confirmed, setting display:none should be deferred. 
Otherwise I think the original implementation was more robust and we may want to re-implement it.

EAlexRojas:
@dmandrioli I performed many tests and the `transitionend` events are always triggered. The only thing that can drop them is to hide the node before the transition finish. 

In the current version, the `transitionend` events are dropped because of the call to the `setVisible()` method on all the nodes. (the for loop on line 319)

I think that each node should set only its own visibility (and "clean" its own CSS classes) depending on whether the node is the `visibleNode` or not...

I also think that we should wait until the two transitions (origin and target) finish to trigger the `delite-after-show` event. 

I'll update this PR with a version that takes all these points into account.

I ran the unit tests on that version using travis and they all pass:
  https://travis-ci.org/EAlexRojas/deliteful/builds/70035372

The test are also passing on the mobile devices.

dmandrioli:
@EAlexRojas Thanks for your tests and explanations. Sounds good to me to update the PR with these points.

dmandrioli:
I think this PR can now be merged.

dmandrioli:
Merged via 0f69918551f943c9cde248aed9b29073d45ef8f4

PR #581, opened 2015-07-14
Fix styling issues when setting the width of the Combobox

gaurav21r:
When we set a CSS `width` property to the root element, the inside `<input>` does not respect its parents width and probably takes browser default. This is a base level issue irrespective of themes.

Incase more info is required, I can attach a screenshot!


wkeese:
It should have a test case, not just a screen shot.

cjolif:
still no test case? If not I guess we will close it.

gaurav21r:
@cjolif Sorry, been busy. Please give me a week's time then close.

PR #627, opened 2016-05-31
Button accessibility fixes / support for icon only button, fixes #624

### LINKED ISSUE #624: Button accessibility fixes
Add support for icon only buttons:

- Allow app to specify a label but have it hidden, for possible benefit of screen readers.  It's unclear if this is really necessary for accessibility but probably doesn't hurt to have it.  See https://www.nczonline.net/blog/2013/04/01/making-accessible-icon-buttons/).
- Set `title` automatically, so that users can hover of (icon only) button and see tooltip explanation?  Add `title={{this.showLabel ? "" : this.label}}` to the template.  OTOH that would be problematic if user specifies a `title` explicitly.
- Fix problem where JAWS announces the button's icon (in addition to the label text) by adding `aria-labelledby` to the button pointing to its text.

brunano21:
- A11y fixes. 
- Add support for icon only buttons (showLabel) and set title accordingly.

Fixes #624

### DIFF
--- a/samples/Button.html
+++ b/samples/Button.html
@@ -1,12 +1,12 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
 	<meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
 	<meta name="viewport"
 		  content="width=device-width,initial-scale=1,maximum-scale=1,minimum-scale=1,user-scalable=no"/>
 	<meta name="apple-mobile-web-app-capable" content="yes"/>
 
-	<title>Button</title>
+	<title id="pageTitle">Button</title>
 
 	<script type="text/javascript" src="../../requirejs/require.js"></script>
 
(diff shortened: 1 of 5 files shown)


samples/Button.html
```diff
@@ -60,17 +60,20 @@
...
-	<button is="d-button" iconClass = 'd-icon-phone d-big-icon' label="Custom with icon" class="d-button-green" onclick="alert(this.label + ' clicked.');"></button>
+	<button is="d-button" iconClass = 'd-icon-phone' label="Custom with icon and label" class="d-button-green" onclick="alert(this.label + ' clicked.');"></button>
+	<button is="d-button" iconClass = 'd-icon-phone d-big-icon' label="Custom with icon only" showLabel="false" class="d-button-green" onclick="alert(this.label + ' clicked.');"></button>
 </div>
 <div class="group">
-	Icon only button (screen readers should announce label as "Call"):
-	<button is="d-button" iconClass="d-icon-phone" title="Call"></button>
+	Icon only button (screen readers should announce label as "Button Call". Hovering with the mouse, a "Call" tooltip will be shown instead.):
+	<br>
+	<button is="d-button" iconClass="d-icon-phone" title="Call" label="Call Button"></button>
```
wkeese: This is labeled as "Icon only button" but it's still showing the button text.   (Conversely, on line 71 there's an icon only button.)

wkeese:
Looks good except for that one issue w/the sample file.

Oh, also you should add "Fixes #624" to the end of the check in comment.

wkeese:
PS: I was going to mention updating Button.md, but strangely it already documents `showLabel`.

wkeese:
Thanks!  I pushed this as 7d023dddb95e0d1def805b959dcac9b31778c2bb.

PR #630, opened 2016-06-16
List/PageableList: refactor. Fixes #626. Fixes #595

### LINKED ISSUE #595: unit test failure in list/PageableList
On 0.8.0 I'm getting this failure:

Update items in displayed page with index page > maxpages

assert 1 number of children: expected 1 to equal 6

My URL is http://localhost/deliteful/node_modules/intern/client.html?config=tests/intern&suites=deliteful/tests/unit/list/PageableList.

### LINKED ISSUE #626: List: "no records" message
Make list [optionally?] display a "no records found" type message when the list is empty.

brunano21:
- List and PageableList now have their own templates.
- Add a customizable no-items node feature
- Hide list containerNode when the list's source is empty.
- Add new loader helper widget for PageableList.
- A11y fixes.
- List and PagebeableList's role can be now set through the `type` property.
- Test fix in order to match the new widgets.
- Update List/PageableList documentation.
- Fixes #626. 
- Fixes #595


list/List.js
```diff
@@ -132,24 +133,47 @@ define([
...
-					// Only role=grid should have aria-readonly.  It's meaningless for role=listbox or role=menu.
-					if (value === "grid") {
-						this.setAttribute("aria-readonly", "true");
-					} else {
-						this.removeAttribute("aria-readonly");
-					}
+		/**
+		 *	If `true` and the list's store is empty, the NoItemContainer element will be shown.
+		 * @member {boolean}
+		 * @default "false"
```
wkeese: Presumably there shouldn't be quotes around "false" because this prop is a boolean.   (There's also some extra space at 137.)

list/List.js
```diff
@@ -132,24 +133,47 @@ define([
...
+		 * It is set to false when the initialization ends.
+		 * Furthermore if `true`, a loading panel will be shown.
+		 * @member {boolean}
+		 * @private
+		 */
+		_busy: false,
+
+		/**
+		 * Specifies the role of list. It can be one of `grid`, `menu`, `listbox`.
+		 * @type {String}
```
wkeese: I think string should be lowercase.

list/List.js
```diff
@@ -132,24 +133,47 @@ define([
...
+
+		/**
+		 * Propertie responsible for showing/hiding the containerNode element, depending of the
+		 * content of the source. It also drives the visibility of d-list-no-items node,
+		 * accordingly to the showNoItems's value.
+		 * @type {String}
+		 * @default ""
+		 * @private
+		 */
+		_displayedPanel: "",
```
wkeese: Can you list the possible values here?  Also, the description is a bit weird; it says that it's responsible for "showing/hiding the containerNode" but that's only one of the things it does.

list/List.js
```diff
@@ -186,6 +210,31 @@ define([
...
+
+		setAttribute: dcl.superCall(function (sup) {
+			return function (name, value) {
+				if (/^aria-/.test(name) && this.containerNode) {
+					this.containerNode.setAttribute(name, value);
+				} else {
+					sup.call(this, name, value);
+				}
+			};
+		}),
```
wkeese: To be thorough you should override the other methods too (getAttribute(), removeAttibute(), hasAttribute()), like FormWidget does.

list/List.js
```diff
@@ -318,22 +364,34 @@ define([
...
+				this._updateListView();
+			}
+		},
+
+		postRender: function () {
+			// moving down to the containerNode any aria attribute that has been set to the root node.
+			for (var i = 0; i < this.attributes.length; i++) {
+				if (/^aria-/.test(this.attributes[i].name)) {
+					this.containerNode.setAttribute(this.attributes[i].name, this.attributes[i].value);
+					this.removeAttribute(this.attributes[i].name);
```
wkeese: If you update to override `removeAttribute()` (as I mentioned above) then this line will need to be changed to

```
// force remove from root node not container node
HTMLElement.prototype.removeAttribute.call(this, this.attributes[i].name);
```

list/PageableList.js
```diff
@@ -265,41 +170,42 @@ define([
...
+		 */
+		_showPreviousPageLoader: false,
+
+		 /**
+		 * Visibility flag for nextPageLoader widget.
+		 * @member {boolean}
+		 * @default false
+		 * @private
+		 */
+		_showNextPageLoader: false,
```
wkeese: There's an inconsistency here with the definition ( "Visibility flag for previousPageLoader widget.") vs. what these flag are actually doing, specifically that sometimes the previous page (and next page) buttons won't be shown even when `_showPreviousPageLoader` and `_showNextPageLoader` are both true.

It's probably better that these flags actually work as they say, so the template can just say:

```
<d-list-loader d-shown={{_showPreviousPageLoader}} ...
```

An alternate approach would be to keep the logic as-is but rename these flags to something more truthful like  `_previousRecordsExist` and `_nextRecordsExist`.

Also the description  would probably be easier to understand if it were something like "Flag for use in template to show/hide the previous page button." (to make it clear what previousPageLoader widget does).

list/List/themes/List_template.less
```diff
@@ -83,69 +83,72 @@
...
+		}
+		.d-list-loading-panel-info-label {
+			padding-left: 10px;
+		}
 	}
 }
 
+.d-list-no-items {
+	white-space: nowrap;
+    text-overflow: ellipsis;
```
wkeese: indentation is off on first line

list/Loader.js
```diff
@@ -0,0 +1,44 @@
...
+define([
+	"dcl/dcl",
+	"delite/register",
+	"delite/Widget",
+	"delite/handlebars!./List/Loader.html"
+], function (dcl, register,	Widget, template) {
+	"use strict";
+
+	/**
+	 * A widget that renders a loader widget.
```
wkeese: Huh?

list/Loader.js
```diff
@@ -0,0 +1,44 @@
...
+		template: template,
+
+		/**
+		 * Set of messages to show for each value of `loading` property.
+		 * @type {Object}
+		 */
+		labels: {
+			loadMessage: "",
+			loadingMessage: ""
+		},
```
wkeese: I was actually expecting just one scalar property to control the button's label, rather than two separate properties.   (As a side note, even if there are two separate properties there's no reason to put them inside a hash called `labels`.)

Although it's debatable, since part of the label setting logic is in PageableList.js, it probably makes sense for all of it to be there.

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
+					this._updateListView();
+				}
+
+				if ("loadPreviousMessage" in props || "loadingMessage" in props || "_showPreviousPageLoader" in props) {
+					if (this.previousPageLoader && this.previousPageLoader.deliver) {
+						this.previousPageLoader.labels = {
+							loadMessage: string.substitute(this.loadPreviousMessage, this),
+							loadingMessage: this.loadingMessage
+						};
+						this.previousPageLoader.deliver();
```
wkeese: `string.substitute(this.loadPreviousMessage, this)` means that loadMessage could change as a response to any property change at all.

So I think probably you should move this logic outside of all the `if()` statements and just say something like:

```
this.previousPageLoader.label = this.previousPageLoader.loading ?
   this.loadingMessage : string.substitute(this.loadPreviousMessage, this);
```

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
+						this.nextPageLoader.deliver();
+					}
+				}
+			}
+		},
+
 		//////////// Private methods ///////////////////////////////////////
 
+		/*
+		 * Ovveride deliteful/List#_updateListview to update the widget view.
```
wkeese: "to update the widget view" is quite vague.  Also, "Override" not "Ovveride".

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
+		 * @private
+		 */
+		_updateListView: function () {
+			this._displayedPanel = (this._busy && this.hideOnPageLoad && !this.autoPaging) ? "loading-panel" :
+				(this.containerNode && this.containerNode.children.length > 0) ?
+					"list" : ((this.showNoItems) ? "no-items" : "none");
+		},
+
+		/*
+		 * Handle click events on the widget.
```
wkeese: The comment is still wrong, should say "next and previous page buttons" not "the widget".

list/PageableList.js
wkeese: PS: Having gone that far, you might as well go the extra step and just define a property `this.PreviousPageButtonLabel` (or something like that), and then let the template reference it, like:

```
<d-list-loader label="{{previousPageButtonLabel}}"></d-list-loader>
```

That's in line with how other template properties are set, and then you don't need to manually call `deliver()` anymore.

list/PageableList.js
wkeese: Oh and also, it's marked as private in both List and PageableList but if it were really private it wouldn't be in PageableList.js at all.   So should be marked as protected, and not have an underscore.

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
+						this.nextPageLoader.labels = {
+							loadMessage: string.substitute(this.loadNextMessage, this),
+							loadingMessage: this.loadingMessage
+						};
+						this.nextPageLoader.deliver();
+					}
+				}
+			}
+		},
+
```
wkeese: Very minor, but can you put computeProperties() before refreshRendering() since that's the order they execute in?

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
+		 * @param {HTML Element} node element
+		 * @return {HTML Element} the loader top node
+		 * @private
+		 */
+		_getLoader: function (node) {
+			while ("d-list-loader" !== node.tagName.toLowerCase()) {
+				node = node.parentElement;
+			}
+			return node;
+		},
```
wkeese: This seems like a lot of code to make _load() work regardless of whether the next or previous button was clicked.  How about having two separate methods _nextPageClickHandler() and _previousPageClickHandler()?

list/PageableList.js
wkeese: PPS: On further review, I think something like this makes the most sense for the template:

```
<d-page-loader loading={{loadingNextPage}} 
label={{this.loadingNextPage ? this.loadingMessage : this. nextPageButtonLabel}}
on-click={{_load}}>
</d-page-loader>
```

The JS code is responsible for updating a `loadingNextPage` boolean property (to reflect whether or not the load is in progress) as well as the `nextPageButtonLabel` string (based on `loadNextMessage` but with substitution variables replaced).

list/PageableList.js
```diff
@@ -315,8 +221,87 @@ define([
...
 
+		/*
+		 * Ovveride deliteful/List#_updateListview to update the widget view.
+		 * @private
+		 */
+		_updateListView: function () {
+			this._displayedPanel = (this._busy && this.hideOnPageLoad && !this.autoPaging) ? "loading-panel" :
+				(this.containerNode && this.containerNode.children.length > 0) ?
+					"list" : ((this.showNoItems) ? "no-items" : "none");
+		},
```
wkeese: Looking at this more, it really seems like code that should be in computeProperties().   (As you know) the only problem is how to trigger that code to run after `itemAdded()`, `itemRemoved()` etc. are called.  It seems like it would be cleaner for `itemAdded()` etc. to call `this.notifyCurrentValue("renderItems")`, and then for computeProperties to have code like:

```
if ("renderItems" in oldVals || "_busy" in oldVals || ... ) {
   this._displayedPanel = ...
}
```

list/PageableList.js
```diff
@@ -461,7 +450,7 @@ define([
 			var renderer = this._getFirstVisibleRenderer();
 			var nextRenderer = renderer.nextElementSibling;
 			if (this.navigatedDescendant) {
-				if (renderer && this._previousPageLoader && this._previousPageLoader.loading) {
+				if (renderer && this._busy) {
```
wkeese: I don't really understand the logic here (either before or after your change) but is there a reason to change the logic to say `this._busy` rather than `this._previousPageLoader.loading`?

list/PageableList.js
```diff
@@ -716,47 +650,36 @@ define([
...
-				}
-				return value;
-			};
-		}),
-
 		_spaceKeydownHandler: dcl.superCall(function (sup) {
 			//	Handle action key on page loaders
 			return function (event) {
-				if (this._nextPageLoader && this._nextPageLoader.contains(event.target)) {
+				if (this._showNextPageLoader && this.nextPageLoader.contains(event.target)) {
```
wkeese: You don't need the `this._showNextPageLoader && ` part of the `if()` statement

list/PageableList.js
```diff
@@ -716,47 +650,36 @@ define([
...
 		handleSelection: dcl.superCall(function (sup) {
 			// page loader should never be selected when clicked
 			return function (event) {
-				var renderer = this.getEnclosingRenderer(event.target);
-				if (renderer === this._nextPageLoader || renderer === this._previousPageLoader) {
+				if (this._showPreviousPageLoader && this.previousPageLoader.contains(event.target) ||
+					this._showNextPageLoader && this.nextPageLoader.contains(event.target)) {
 					return;
 				} else {
 					sup.apply(this, arguments);
```
wkeese: I don't think you need this method at all.

Rather, shouldn't List.js be changed so instead of:

```
this.on("click", this.handleSelection.bind(this));
```

it does that for `this.containerNode`?   You could put `on-click={{_handleSelection}}` in the template.

Having said that, I don't know how keyboard selection is working, so you should investigate that.

tests/unit/list/PageableList.js
```diff
@@ -1235,14 +1224,61 @@ define([
...
+			list = new PageableList({source: new Store()});
+			for (var i = 0; i < 92; i++) {
+				list.source.add({label: "item " + i});
+			}
+			list.hideOnPageLoad = true;
+			list.pageLength = 20;
+			list.maxPages = 0;
+			document.body.appendChild(list);
+			list.attachedCallback();
+			list._busy = true;
```
wkeese: Is this line necessary?  It seems weird.

wkeese:
It's looking better now.  The functional tests are hanging though, since they are searching for aria-busy on the wrong element.

wkeese:
I'm testing this against various browsers, which is difficult since there are pre-existing failures and also since testing against Saucelabs gets intermittent failures (not due to problems with deliteful AFAICT).  But for IE11/Win8, I see the following **new** failures on this branch:

```
× internet explorer 11 on Windows 8.1 - unit tests - list/PageableList - list in a popup (0.075s)
AssertionError: list.offsetWidth === 2
No stack or location
× internet explorer 11 on Windows 8.1 - unit tests - list/PageableList - list with categories in a popup (0.116s)
AssertionError: list.offsetWidth === 2
No stack or location
× internet explorer 11 on Windows 8.1 - unit tests - list/PageableList - Add item after last page creates loader (0.076s)
AssertionError: 0: last children is next page loader: expected '' to equal 'Click to load 23 more items'
No stack or location
× internet explorer 11 on Windows 8.1 - unit tests - list/PageableList-ObservableArray-noPagination-baseListTests - first focus apply to the first visible child (0.034s)
AssertionError: active element: expected null to not equal null
No stack or location
× internet explorer 11 on Windows 8.1 - unit tests - list/PageableList-ObservableArray-noPagination-baseListTests - show/hide no items node depending of list's containerNode children (0.023s)
AssertionError: .d-list-no-items must be visible: expected null to not equal null
No stack or location
× internet explorer 11 on Windows 8.1 - Pageable tests - page loaders cannot be selected (4.641s)
ElementNotVisible: [POST http://(redacted)@localhost:4444/wd/hub/session/83cbd074a1b44652a0f019dd689c1c11/element/539/click]
```

list/List.js
```diff
@@ -182,6 +182,8 @@ define([
 
 		template: template,
 
+		tabindex: 0,
```
wkeese: The property is actually `tabIndex` with a capital I.  In theory you should be able to add that property and then it works, but you'll have to see if you get problems on Chrome.

brunano21: I have renamed the `tabindex` property with `tabIndex` but this generate errors while parsing the template. I guess because tabIndex is reserved keyword.

brunano21: Potentially I may rename the `tabIndex` property into something else, like `_tabIndexAttr` and bind it into the template. Makes sense to you?

wkeese: No, that shouldn't be necessary.  See delite/tests/functional/TabIndex.html.

wkeese:
OK, I finally pushed this!   Checked in as 3338825d03ab7b66152fd7bcae9d30d9428c2fff.

I also did a bunch of minor corrections in 2610df165fdc07372f5ede45e03306b3bdf7a804, plus it turned out there was another delite change needed too (ibm-js/delite@936e33dafe162919e1c13dbbeb1cbf4c922b856c in addition to ibm-js/delite@4412cc3e00695d3fad347d09c6623e1150614c11 and ibm-js/delite@41d25cc5e4f77f03333b55c3e4d052e7d00915f9).

PR #634, opened 2016-06-29
Fixes #619

### LINKED ISSUE #619: List: support role=list
When currently support role=grid (strangely the default) and role=listbox and role=menu, but not role=list.

brunano21:
This PR adds support for the aria role list. Unit tests included.


wkeese:
Thanks, pushed as c1766c27fa64deb3b90b18d5762c451cade28721.  (I just had to update the checkin comment to include "Fixes #619".

PR #637, opened 2016-07-22
Combobox accessibility fixes

### LINKED ISSUE #636: Combobox: accessibility broken on mobile
On iOS, you can't use VoiceOver with Combobox.  In particular, the problem is that after the popup opens, it sends focus back to the original Combobox <input>, behind the popup.

Bruno has a patch that addresses at least some of the accessibility issues.

wkeese:
Fixes #636.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -362,11 +362,7 @@ define([
 
 			// The role=listbox is required for the list part of a combobox by the
 			// aria spec of role=combobox
-			this.list.setAttribute("role", "listbox");
-
-			// Avoid that List gives focus to list items when navigating, which would
-			// blur the input field used for entering the filtering criteria.
-			this.list.focusDescendants = false;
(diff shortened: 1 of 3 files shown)


wkeese:
Thanks for the patch.   I'm pushing this with a few minor tweaks:

1. jshint spacing fix
2. I decided to modify delite/Dialog to search for tabbable items in `this` rather than `this.containerNode`.   ISTM that if for some reason the Dialog chrome has navigable items, then the user should be able to navigate to them.   Having said that, if the Dialog has a Ⓧ button, there's no reason to make it tab navigable, except maybe for mobile, because on desktop the user can cancel the dialog with the ESC key.
3. squashed commits

wkeese:
Pushed as 17bb3e27532b2c3e440928a69e112ef6ddbd3c61.

PR #638, opened 2016-07-25
Close/open combobox's popup while a filtering operation is in progress.

brunano21:
This PR closes and opens the dropdown before and right after a filtering operation has performed, allowing so to calculate again size and position of the popup wrapper. 

Moreover, it updates samples and tests page in order to leverage on List's `showNoItems` feature, when the query returns an empty set. It may be worth to set this feature by default. Of course it makes sense only when `autoFilter=true`.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -576,7 +576,7 @@ define([
 		},
 
 		/**
-		 * Indicates if a filterin operation is in progress. If so, closeDropDown will perform differently than usual.
+		 * Indicates if a filtering operation is in progress. If so, closeDropDown will perform differently than usual.
 		 * @member {boolean}
 		 * @default false
 		 */
--- a/Combobox.js
+++ b/Combobox.js
@@ -589,13 +589,13 @@ define([
 				// events, triggered when pressing ENTER. This would also fit for Chrome/Android,
 				// where pressing the search key of the virtual keyboard also triggers a
 				// change event. But there's no equivalent on Safari / iOS...
-				if(this.opened && !this.filteringInProgress) {
+				if (this.opened && !this.filteringInProgress) {
 					this.filteringInProgress = true;
 					this.closeDropDown();
 				}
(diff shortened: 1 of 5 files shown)


wkeese:
I guess this is good.  We should add an automated test though, to make sure the popup stays aligned and visible as it's filtered

Combobox.js
```diff
@@ -577,15 +577,33 @@ define([
 			return new ComboPopup({combobox: this});
 		},
 
+		/**
+		 * Indicates if a filterin operation is in progress. If so, closeDropDown will perform differently than usual.
```
wkeese: filterin --> filtering

brunano21:
@wkeese "We should add an automated test though, to make sure the popup stays aligned and visible as it's filtered" how do you think we can verify this? I guess checking the popup's visible isn't enough.

wkeese:
Right, I thought something like the delite/place test does:

```
aroundB: function () {
	// Dropdown from "aroundBottom" node. Should go above aroundNode so that
	// popup doesn't go off screen.
	scrollTo(100);
	var ret = place.around(popup, aroundBottom, [
		"below",	// aroundBottom's bottom-left corner with the popup's top-left corner (fails)
		"above",	// aroundBottom's top-left corner with the popup's bottom-left corner (works)
		"below-alt"	// aroundBottom's bottom-right corner with the popup's top-right corner (fails)
	], true);

	assert.strictEqual(ret.aroundCorner, "TL", "around corner");
	assert.strictEqual(ret.corner, "BL", "popup's corner");

	var aroundPos = aroundBottom.getBoundingClientRect(),
		popupPos = popup.getBoundingClientRect();
	assert.strictEqual(Math.round(aroundPos.top), Math.round(popupPos.bottom), "above around node");
},
```

wkeese:
Thanks, pushed as 3d8883e76d2b21517b439865434cf3051d2cce6e with further simplification to sample file.

PR #641, opened 2016-07-27
Toggle Button: fix a11y issues

### LINKED ISSUE #640: ToggleButton accessibility issues.
Add support for icon-only ToggleButtons:
 - Allow app to specify a label but have it hidden, for possible benefit of screen readers, leveraging on `deliteful/Button`'s `showLabel` property.
 - Set title automatically, so that users can hover of (icon only) button and see tooltip.
 - Fix problem where JAWS announces the button's icon (in addition to the label text) by adding `aria-label` attribute to the widget's `rootNode`.

brunano21:
- Add tests.
- Update Doc. 
- Fixes #640

### DIFF
--- a/ToggleButton.js
+++ b/ToggleButton.js
@@ -22,6 +22,10 @@ define([
 	 * specified by the 'iconClass' property is applied whatever the state is. However, a css class specific to the
 	 * checked state can be specified via the 'checkedIconClass' property.
 	 *
+	 * Moreover, a toggle button can show an icon only with no visible text, independently of
+	 * the toggle button's state, checked or unchecked. To accomplish that,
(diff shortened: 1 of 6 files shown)


wkeese:
This patch looks good.  The only minor thing I noticed is that all the asserts are backwards, they should be `assert.strictEquals(actualVal, expectedVal)` rather than  `assert.strictEquals(expectedVal, actualVal)` so that if the assert fails it prints an error message like "got 123 but expected 456".

brunano21:
Uhm, I guess I got mislead from what I have seen in the preexisting tests (of the same file). 
I'll fix that soon.

wkeese:
Thanks, pushed as 9e87106a517ea3c9473fa8ed00d4de6a2f6ff0cc.

Issue #639, opened 2016-07-26
Combobox: detached popup while filtering.


brunano21:
When a `d-combobox`'s list gets filtered and the popup is shown, the position and size of the popup does not get adjusted accordingly. 
An example is when the popup is above the `d-combobox` widget. In fact, in this scenario if the user types something into the input field and the list shows only matching items, then the popup seems to detached from the `d-combobox` widget itself.

PR #644, opened 2016-08-09
Fix Combopopup on mobile

wkeese:
Fix regression in Combobox on mobile where typing into the `<input>` in the Comopopup makes the whole popup close.

Refs #639.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -591,15 +591,17 @@ define([
 				// change event. But there's no equivalent on Safari / iOS...
 				if (this.opened && !this.filteringInProgress) {
 					this.filteringInProgress = true;
-					this.closeDropDown();
+					this.closeDropDown(true);
 				}
 				// this.filter() call will fire a query-success event. After that, the popup can be opened again.
-				this.own(this.list.on("query-success", function () {
+				var cb = this.own(this.list.on("query-success", function (evt) {
+					console.log(evt.renderItems);
 					if (this.filteringInProgress) {
 						this.filteringInProgress = false;
+						cb.remove()
 						this.openDropDown();
 					}
-				}.bind(this)));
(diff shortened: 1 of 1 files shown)


wkeese:
@brunano21, this definitely fixes the problem, but I found another issue with the original fix for #639...

I tracked down why the `<input>` is cleared, and what's actually happening is that every time you call `closeDropDown()` and `openDropDown()`, the dropdown is being **recreated**`.  It's because `HasDropDown#openDropDown()` calls `Combobox#loadDropDown()`, which unconditionally calls `createCenteredDropDown()`, which creates a new `ComboPopup`.   Seems like it would be better (faster) if we could avoid doing that.  

Looks like desktop doesn't have that problem because although `Combobox#loadDropDown()` unconditionally calls `createAboveBelowDropDown()`, that method simply returns `this.list`.

wkeese:
PS: Another problem with the current design is that the dialog actually jumps around as you type.  But I guess one problem at a time.

brunano21:
@wkeese Ok, I see your point. However I think the original problem of this (closing/opening the popup) is due to the lack of a `redraw()` method in `delite/popup.js`. As matter of fact, we tend to close and open the `dropDown` because we want to recalculate the new `dropDown`'s size when its content changes.
How worth is it to implement a `redraw()` method? is that possible? It would make things much easier _I think_.

wkeese:
Agreed, I'll try to add a `redraw()` method.  Actually I think it's already there in the `_repositionAll()` method so it's just a question of making that or a similar method public.

brunano21:
The `delite/popup#_repositionAll()` isn't accessible, because `delite/HasDropDown` doesn't inherit from `delite/popup`.

wkeese:
Replaced by #648.

PR #647, opened 2016-08-18
Fixes #646

### LINKED ISSUE #646: Accordion: accessibility violation.
At this stage, each header has `aria-controls` set even though its panel is not visible.
This represents an accessibility violation, as stated below:

> **Elements that use WAI-ARIA properties must reference valid IDs**
> Elements that use WAI-ARIA properties must reference valid IDs to enable compatibility with assistive technologies.
> Following is a list of invalid ID references:
> 
>    1. Element referenced by ID doesn't exist.
>    2. WAI-ARIA properties only supports single ID reference but multiple provided.
>    3. Empty or whitespace ID reference.
>    4. _Element referenced by ID is hidden._ 

`aria-controls` attribute should be set only when the panel is visible and should be removed once the panel is hidden.

brunano21:
- Fixes #646 
- Fixes few a11y violation in the Accordion sample page.

### DIFF
--- a/Accordion.js
+++ b/Accordion.js
@@ -176,7 +176,6 @@ define([
 			header.setAttribute("role", "tab");
 			header.setAttribute("aria-expanded", "false");
 			header.setAttribute("aria-selected", "false");
-			header.setAttribute("aria-controls", panel.id);
 			header.on("click", this._headerClickHandler.bind(this));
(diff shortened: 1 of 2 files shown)


wkeese:
Thanks, pushed as a156e5578a70ee037f1fb6278b79ee0c42cfecbc.

PR #648, opened 2016-08-18
Combobox popup redraw

### LINKED ISSUE #639: Combobox: detached popup while filtering.

wkeese:
Redo how the popup is repositioned when its size changes, depends on delite PR change to popup.

Fixes #639 again.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -273,8 +273,8 @@ define([
 				this.list = new List();
 			}
 
-			this.list.own(this.list.on("query-success", function (evt) {
-				if(this.opened && this.filteringInProgress) {
+			this.list.own(this.list.on("query-success", function () {
+				if (this.opened && this.filteringInProgress) {
 					this.filteringInProgress = false;
 					this.defer(function () {
 						popup.redraw(this);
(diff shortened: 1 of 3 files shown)


Combobox.js
```diff
@@ -8,6 +8,7 @@ define([
 	"delite/CssState",
 	"delite/FormValueWidget",
 	"delite/HasDropDown",
+	"delite/popup", // redraw()
```
wkeese: You don't need this anymore, right?

Combobox.js
```diff
@@ -589,18 +591,8 @@ define([
...
-				}
-				// this.filter() call will fire a query-success event. After that, the popup can be opened again.
-				this.own(this.list.on("query-success", function () {
-					if (this.filteringInProgress) {
-						this.filteringInProgress = false;
-						this.openDropDown();
-					}
-				}.bind(this)));
 
+				// this.filter() call will fire a query-success event. After that, the popup can be opened again.
```
wkeese: And this comment is old too?

wkeese:
Pushed as 245ba117c9134e71d6037d2316a3edfcd9ca0c7f.

PR #651, opened 2016-08-31
Fixes #650

### LINKED ISSUE #650: Toggle Button: erroneous type.
Currently the ToggleButton's `type` is `submit`, which could cause some problems while checking/unchecking.
Its `type` should be `button` instead.

brunano21:
Fixes #650


wkeese:
Thanks, pushed as 5194f6cfa5dc29b238934ff3291a60ecfe695a4b.

Issue #653, opened 2016-09-18
List should emit `delite-size-change` every time its content changes.


brunano21:
With the current implementation, `List` emits a `delite-size-change` event when `this.renderItems` changes. However this is not enough and does not cover all scenarios. 
In fact, the event should be emitted every time the list's content changes, including when displaying the NoItems node and the LoadingPanel node.

`List#refreshRendering()` should be rearranged like below (`deliteful/list/List.js`):

```
refreshRendering(props) {
    ...
    if (("renderItems" in props && this.renderItems) ||
        ("_displayedPanel" in props && this._displayedPanel !== props._displayedPanel)) {
        // notify the list content changed.
        this.emit("delite-size-change");
    }
    ...
}
```

wkeese:
Agreed that it should also emit that event when the displayed panel changes.

Checking for things like `this._displayedPanel !== props._displayedPanel` is a bit weird since the whole point of `props` is to only list the properties that have changed value.  Is the `this._displayedPanel !== props._displayedPanel` code to avoid the special case on widget initialization when `refreshRendering()` is called for every property?   If so, maybe using the `isAfterInitialRendering` parameter makes more sense:

```js
/**
 * Callback function to render UI upon property changes.
 * @param {Object} oldValues The hash table of old property values, keyed by property names.
 * @param {boolean} isAfterInitialRendering True if this call is right after `initializeRendering()`.
 */
refreshRendering: function () {}
```

It amounts to the same thing; just a style issue.

brunano21:
Actually the check I did was meant to be in that way, `this._displayedPanel !== props._displayedPanel`. In fact as you said, `props` brings all those properties that have changed. If `_displayedPanel` value changed, this means that the popup is going to contains something that could not fit properly, so the `delite-size-change` event should be fired.

Or do you see a different way to know when the `_displayedPanel` value changed?

wkeese:
How about `"_displayedPanel" in props && !isAfterInitialRendering`?   Sounds like you don't think that works.  What case do you think it fails in?

PR #656, opened 2016-09-27
Combobox refactor

wkeese:
This is Bruno's WIP to refactor the Combobox to be friendly for server-side stores.   Also adding the hasDropDownArrow flag so it can look like an autocomplete widget rather than a Combobox.

I'm just creating the PR so it's easier to see the diff.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -267,13 +267,6 @@ define([
 		 */
 		openOnPointerDown: true,
 
-		/**
-		 * When using a ComboPopup (hence in mobile), this properties specifies which node gets the focus
-		 * once the popup is open.
-		 * Possibile values are "input" and "list".
-		 */
-		nodeToFocus: "listNode",
-
 		createdCallback: function () {
 			// Declarative case (list specified declaratively inside the declarative Combobox)
 			var list = this.querySelector("d-list");
@@ -606,9 +599,7 @@ define([
 		 * @protected
 		 */
 		createCenteredDropDown: function () {
-			var node = (this.selectionMode === "single" && this.autoFilter
-				&& this.nodeToFocus !== "list" ) ?  this.nodeToFocus : "list";
-			return new ComboPopup({combobox: this, nodeToFocus: node});
+			return new ComboPopup({combobox: this});
 		},
 
 		/**
--- a/Combobox.js
+++ b/Combobox.js
@@ -252,6 +252,28 @@ define([
 		 */
 		cancelMsg: messages["cancel-button-label"],
 
+		/**
+		 * Displays or not the down arrow button.
+		 * @type {Boolean}
+		 * @default true
+		 */
+		hasDownArrow: true,
+
+		/**
+		 * Defines if a click on widget itself will show up the list immediately.
+		 * It makes sense only on desktop version, since in mobile, a click on the widget opens the
+		 * popup widget always.
+		 * @type {Boolean}
+		 */
+		openOnPointerDown: true,
+
+		/**
+		 * When using a ComboPopup (hence in mobile), this properties specifies which node gets the focus
+		 * once the popup is open.
+		 * Possibile values are "input" and "list".
+		 */
+		nodeToFocus: "listNode",
+
 		createdCallback: function () {
 			// Declarative case (list specified declaratively inside the declarative Combobox)
 			var list = this.querySelector("d-list");
@@ -420,7 +442,19 @@ define([
 			}.bind(this));
 
 			this._prepareInput(this.inputNode);
+
+			if (this._useCenteredDropDown()) {
+				this.on("click", this.openDropDown.bind(this), this);
+			} else if (this.openOnPointerDown) {
+			 	this.on("mousedown", this.openDropDown.bind(this), this);
+			}
 		},
+
+		/**
+		 * Defines if the input value has to be filled with first List's item option.
+		 * @type {Boolean}
+		 */
+		preSetValue: true,
 
 		/**
 		 * Sets the initial value of the widget. If the widget is inside a form,
@@ -457,7 +491,7 @@ define([
 					return done;
 				}.bind(this);
 
-				if (!initValueSingleMode()) {
+				if (this.preSetValue && !initValueSingleMode()) {
 					// List not ready, wait.
 					// TODO: handle case when List is initialized but has no items yet, from "new List()".
 					// Also, handle when the Combobox's selected item is deleted from the list.
@@ -572,15 +606,23 @@ define([
 		 * @protected
 		 */
 		createCenteredDropDown: function () {
-			return new ComboPopup({combobox: this});
+			var node = (this.selectionMode === "single" && this.autoFilter
+				&& this.nodeToFocus !== "list" ) ?  this.nodeToFocus : "list";
+			return new ComboPopup({combobox: this, nodeToFocus: node});
 		},
 
 		/**
-		 * Indicates if a filtering operation is in progress. If so, closeDropDown will perform differently than usual.
-		 * @member {boolean}
-		 * @default false
+		 * Timeout handle for interrupting previouse .
+		 * @type {Number}
 		 */
-		filteringInProgress: false,
+		_timeoutHandle: null,
+
+		/**
+		 * Defines the milliseconds the widget has to wait until a new filter operation starts.
+		 * @type {Number}
+		 * @default 0
+		 */
+		filterDelay: 0,
 
 		_prepareInput: function (inputElement) {
 			this.on("input", function (evt) {
@@ -589,8 +631,12 @@ define([
 				// events, triggered when pressing ENTER. This would also fit for Chrome/Android,
 				// where pressing the search key of the virtual keyboard also triggers a
 				// change event. But there's no equivalent on Safari / iOS...
-
-				this.filter(inputElement.value);
+				if (this._timeoutHandle !== null) {
+					this._timeoutHandle.remove();
+				}
+				this._timeoutHandle = this.defer(function () {
+					this.filter(inputElement.value);
+				}.bind(this), this.filterDelay);
 				// Stop the spurious "input" events emitted while the user types
 				// such that only the "input" events emitted via FormValueWidget.handleOnInput()
 				// bubble to widget's root node.
@@ -691,6 +737,10 @@ define([
 			} // nothing to add for "contains"
 
 			var rexExp = new RegExp(filterTxt, this.ignoreCase ? "i" : "");
+			var queryListener = this.list.on("query-success", function () {
+					this.openDropDown();
+					queryListener.remove(queryListener);
+			}.bind(this));
 			this.list.query = (new Filter()).match(this.list.labelAttr, rexExp);
 		},
 
@@ -722,9 +772,12 @@ define([
 					// Avoid that List gives focus to list items when navigating, which would
 					// blur the input field used for entering the filtering criteria.
 					this.dropDown.focusDescendants = false;
-
-					this._updateScroll(undefined, true);	// sets this.list.navigatedDescendant
-					this._setActiveDescendant(this.list.navigatedDescendant);
+					if (!this._useCenteredDropDown()) {
+						this._updateScroll(undefined, true);	// sets this.list.navigatedDescendant
+						this._setActiveDescendant(this.list.navigatedDescendant);
+					} else {
+						this.dropDown.focus();
+					}
 				}.bind(this));
 			};
 		}),
--- a/Combobox.js
+++ b/Combobox.js
@@ -280,10 +280,11 @@ define([
 		source: null,
 
 		/**
-		 * Flag to initialize initial widget's value, inputValue's value and valueNode's value.
-		 * @type {Boolean}
+		 * Minimum number of characthers before a filter operation runs.
+		 * @type {Number}
+		 * @default 1
 		 */
-		_initValues: true,
+		minChars: 1,
 
 		/**
 		 * Initial inputNode's value.
@@ -291,17 +292,14 @@ define([
 		 */
 		displayedValue: "",
 
-		/**
-		 * This flags allow to force the selection of list's items.
-		 * It needs to be set to true right after widget.value has been set with selected items of the list.
-		 * @type {Boolean}
-		 */
-		setSelectedItems: false,
-
 		createdCallback: function () {
 			// Declarative case (list specified declaratively inside the declarative Combobox)
 			var list = this.querySelector("d-list");
 			if (list) {
+				if (list.source !== null) {
+					// The widget has to get the list's source, if any.
+					this.source = list.source;
+				}
 				if (!list.attached) {
 					list.addEventListener("customelement-attached", this._attachedlistener = function () {
 						list.removeEventListener("customelement-attached", this._attachedlistener);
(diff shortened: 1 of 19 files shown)


tests/functional/SlowStore.js
```diff
@@ -0,0 +1,69 @@
...
+				} else if (query.filterMode === "is") {
+					filterTxt = "^" + filterTxt + "$";
+				}
+				var regExp = new RegExp(filterTxt, query.ignoreCase ? "i" : "");
+
+				for (var i = 0; i < this.items.length; i++) {
+					if (this.items[i].label.match(regExp)) {
+						matchingItems.push(this.items[i]);
+					}
+				}
```
wkeese: match() is for when you need to see the exact results of how it matched.   If you are just checking if it matched or not you should use test().   Combine that w/using the Array methods and you can replace this for() loop with:

```
matchingItems = this.items.filter(regExp.test.bind(regExp))
```

Combobox.js
```diff
@@ -252,10 +252,54 @@ define([
...
+		displayedValue: "",
+
 		createdCallback: function () {
 			// Declarative case (list specified declaratively inside the declarative Combobox)
 			var list = this.querySelector("d-list");
 			if (list) {
+				if (list.source !== null) {
+					// The widget has to get the list's source, if any.
+					this.source = list.source;
+				}
```
brunano21: We actually need that code because this is the case when the developers create a Combobox decoratively, like below:
```
<d-combobox selectionMode="multiple" id="comboPlayersWithValuesDecl" autoFilter="true" value="Hagi,Buffon">
	<d-list>
		{ "label": "Zidane" },
		{ "label": "Platini" },
		{ "label": "Maradona" },
		{ "label": "Hagi" },
		{ "label": "Buffon" }
	</d-list>
</d-combobox>
```
In this case we need to catch the list's source, so the widget has his own copy. 

Aside that we don't need to nullify the `list.source`. In fact even though the list is already initialized, it won't fire any spurious request, because the source is just a Array store (and not i.e. a `dstore/Request`).

Combobox.js
```diff
@@ -767,6 +883,28 @@ define([
...
+			return function (name, value) {
+				if (/^aria-/.test(name)) {
+					this.forEachFocusNode(function (node) {
+						node.setAttribute(name, value);
+					});
+				} else {
+					sup.call(this, name, value);
+				}
+			};
+		}),
```
brunano21: Actually I don't really know where this came from. :confused:

Combobox.js
wkeese: Why do you think the Combobox should have its own copy?    What the point of all the `if (this.source)` conditions in other parts of Combobox.js?

brunano21: Ok maybe I'm not following you anymore. So are you saying that we don't need anymore that block (lines 299-302) AND all those `if  (this.source)` conditions, assigning so the combo's source to the list (no matter of its value)?

wkeese: No, I'm saying we don't need that block (lines 299-302) _because of_ all those `if (this.source)` conditions.

Combobox.js
```diff
@@ -394,89 +434,45 @@ define([
...
+				}.bind(this));
+			} else if (!this.minFilterChars) {
+				this.on("mousedown", function (evt) {
+					// event could be triggered by the down arrow. If so, we do not react to it.
+					if (!this.disabled && evt.srcElement !== this.buttonNode) {
+						if (!this.opened) {
+							this.openDropDown(false);
+						} else {
+							this.closeDropDown(true);
+						}
```
wkeese: I'm a bit worried about closing the dropdown due to a click on the `<input>` (as opposed to `this.buttonNode`), because the user might click the `<input>` just to move the caret.   What do you think?

Combobox.js
```diff
@@ -575,22 +571,50 @@ define([
 			return new ComboPopup({combobox: this});
 		},
 
+		evaluateFilteringConditions: function (inputElement) {
```
wkeese: Can you add a comment explaining what this method does?   From context it seems to be checking whether or not the Combobox should query the store and show the dropdown, but should make that explicit.

wkeese: Oh, also, if that's what the function does, then maybe a better name would be "shouldRunQuery" or "shouldShowDropDown" or something like that?   The current name confused me because the method isn't actually running the filter.

Combobox.js
```diff
@@ -575,22 +571,50 @@ define([
...
+			var mobile = this._useCenteredDropDown();
+			if (this.minFilterChars !== inputElement.value.length !== 0 &&
```
wkeese: The code seems to work but I'm confused about this condition for a number of reasons.  I wonder if you could make it clearer.

1. `this.minFilterChars !== inputElement.value.length`: It seems like you should be testing for `minFilterChars > inputElement.value.length` or something like that.   Why the `!==`?  In other words, what's special about the case where `minFilterChars` and `inputElement.value.length` are exactly the same?
2. I don't how a condition with two `!==` works.  For example, if `minFilterChars` is 0 and `inputElement.value.length` is 1, the condition becomes `0 !== 1 !== 0`, but is that true or false?  It's quite unusual to write a condition like this so maybe you can rewrite using an `&&`?

Combobox.js
```diff
@@ -641,65 +665,145 @@ define([
...
-		 * The default implementation uses `dstore/Filter.match()`.
-		 * The matching is performed against the `list.labelAttr` attribute of
-		 * the data store items.
-		 * The method can be overridden for implementing other filtering strategies.
 		 * @protected
 		 */
-		filter: function (filterTxt) {
-			if (this.filterMode === "startsWith") {
+		filter: function (inputText) {
+			// Escape special chars in search string, see
```
wkeese: Minor thing but indentation on this line is off.

Combobox.js
```diff
@@ -641,65 +665,145 @@ define([
...
+		/**
+		 * Sets the new list's query.
+		 * This method can be overridden when using other store types.
+		 * The default implementation uses `dstore/Filter.match()`.
+		 * The matching is performed against the `list.labelAttr` or `list.labelFunc` attributes of
+		 * the data store items.
+		 * The method can be overridden for implementing other filtering strategies.
+		 * @protected
+		 * @returns {Object} New query to set to the list.
+		 */
```
wkeese: Minor thing but you should document what the `args` parameter is.

Combobox.js
```diff
@@ -641,65 +665,145 @@ define([
...
+
+		/**
+		 * Sets the initial list's query,
+		 * It can be one of a `function`, an `Object` or `null`.
+		 * If it is a function, then it's invoked and the list's query will get the return value assigned.
+		 * If it is an Object, it's assigned to the list's query directly.
+		 * If it is null, the list's query won't be set to its initial value.
+		 * It can be overridden depending of store used and the strategy to apply.
+		 */
+		getInitialQuery: {},
```
wkeese: I think we talked about this before, but I don't remember what "initial" means here.  Is it the query used when there's no text in the `<input>`?

Combobox.js
```diff
@@ -767,6 +875,22 @@ define([
 			};
 		}),
 
+		// HasDropDown#_dropDownKeyUpHandler() override.
+		// Do not call openDropDown if widget does not have a down arrow shown.
+		_dropDownKeyUpHandler: dcl.superCall(function (sup) {
+			return function () {
+				if (this.hasDownArrow) {
+					if (this.inputNode.value.length === this.minFilterChars === 0) {
+						this.list.query = this.getInitialQuery;
```
wkeese: This must be wrong because you aren't accounting for `getInitialQuery` being a function.  In other parts of the code you do account for that.

Combobox.js
```diff
@@ -641,65 +665,145 @@ define([
...
-			return function () {
-				// Store the current selection, to be able to restore when pressing the
-				// cancel button. Used by ComboPopup. (Could spare it in situations when
-				// there is no cancel button, but not really worth.)
-				this._selectedItems = this.list.selectedItems;
+			return function (filteringInProgress) {
+				// Assign widget's source to the list's source.
+				if (this.source) {
+					this.list.source = this.source;
+				}
```
wkeese: I think we talked about always setting `this.list.query` and `this.list.source` at the same time (or in other words, defer setting `this.list.source` until you set `this.list.query`.   Maybe that's what's happening here, but it's hard to tell because I just see the line for `this.list.source`.

It would be clearer if you always put those two together.

Maybe you want a helper function to do that?  It could also handle `getInitialQuery`, specifically the part about branching depending on if `getInitialQuery` is an Object or a function.

Combobox.js
```diff
@@ -767,6 +875,22 @@ define([
...
+					if (this.inputNode.value.length === this.minFilterChars === 0) {
+						this.list.query = this.getInitialQuery;
+					} else if (this.inputNode.value.length < this.minFilterChars) {
+						return;
+					}
+
+					sup.call(this);
+				}
+			};
+		}),
```
wkeese: Looks like Combobox-decl.html is missing a test for when hasDownArrow == false?   Then by corollary, you are missing an automated test for this piece of code?

I'm also confused by what this code is doing.  The comment says "Do not call openDropDown if widget does not have a down arrow shown.", but of course the dropdown is supposed to open at some point.   Can you make the method description clearer?

Combobox.js
```diff
@@ -722,19 +826,29 @@ define([
...
+						this._updateScroll(undefined, true);	// sets this.list.navigatedDescendant
+						this._setActiveDescendant(this.list.navigatedDescendant);
+					} else {
+						// mobile version
+						if (!this.hasDownArrow) {
+							if (filteringInProgress && this.inputNode.value.length >= this.minFilterChars) {
+								$(this.list).removeClass("d-combobox-list-hidden");
+							} else {
+								$(this.list).addClass("d-combobox-list-hidden");
+							}
```
wkeese: You could use the `d-hidden`/`d-shown` class or alternately set the `d-hidden`/`d-shown` attribute, rather than creating a new `d-combobox-list-hidden` class.   So this code could be changed to:

```js
this.list.setAttribute("d-shown", filteringInProgress &&
    this.inputNode.value.length >= this.minFilterChars);
```

Similarly of course for other code accessing `d-combobox-list-hidden`.

list/List.js
```diff
@@ -368,7 +368,8 @@ define([
 				}
 			}
 
-			if ("renderItems" in props && this.renderItems) {
+			if (("renderItems" in props && this.renderItems) ||
+				("_displayedPanel" in props && this._displayedPanel !== props._displayedPanel)) {
```
wkeese: I see this is the fix for #653, but see my comment in that ticket.

Combobox.js
brunano21: Yeah I didn't think about this edge case. I guess that the _expected behavior_ could be to not close the popup if the user clicks into the input text. Of course, this makes sense only if the input **is not** in read-only mode.

Combobox.js
```diff
@@ -272,7 +305,8 @@ define([
 			}
 		},
 
-		refreshRendering: function (oldValues) {
+		/* jshint maxcomplexity: 17 */
+		refreshRendering: function (oldValues, justCreated) {
 			var updateReadOnly = false;
 			if ("list" in oldValues) {
 				this._initList();
```
brunano21: Don't you think we should use a flag to prevent multiple `_initList()` calls? This could cause the binding of multiple event listener. See `_prepareInput()` please.

Combobox.js
brunano21: That's exactly what I meant when I was saying to be careful during the review. In fact I'm not fully convinced neither and I think the function could be rewritten in a better/elegant way.

Combobox.js
brunano21: I don't see where.

Combobox.js
brunano21: I would define "initial" as the default query of the widget. I.e. in a weird(?!) scenario we could set it in order to filter by default all items starting with `a` char.

Combobox.js
brunano21: Sure. I'll do. Basically, if the widget does not have an down arrow, this means the widget is configured in auto-complete mode. In this mode we cannot open the dropdown; it will be opened once the user typed sometime, with `something.length > this.minFilterChars`.

Combobox/ComboPopup.js
```diff
@@ -36,6 +36,12 @@ define([
 		 */
 		header: "",
 
+		/**
+		 * Specifies the visibility of the inner list
+		 * @type {boolean}
+		 */
+		showList: true,
```
brunano21: At line 54 you can see it's used instead.

samples/Combobox.html
```diff
@@ -41,6 +43,9 @@
 				comboTeams.ignoreCase = ignoreCase.checked;
 			});
 
+			$("input[type=number").on('keyup change input', function(evt) {
```
brunano21: 1. Yeah.
2. Maybe yes.
3. Sorry about that.

Combobox.js
wkeese: This line:

```
// Escape special chars in search string, see
```

is indented more than:

```
// http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex.
```

and subsequent lines.  It might also be a problem due to using spaces instead of tabs.

Combobox.js
wkeese: It seems like a trick question... we only want to call `_initList()` once for a list, but [in the unlikely event that] `this.list` changes then we should call `_initList()` again.

Probably though line 311 should read `if ("list" in oldValues && this.list)` to prevent calling `_initList()` when `this.list` is null.

I did look at `prepareInput()` but I don't understand the connection.

Combobox.js
wkeese: Yeah I guess so too.  It probably doesn't matter much; like you said, it's just an edge case.

Combobox.js
wkeese: 👍

Combobox.js
wkeese: Right... so maybe `getDefaultQuery()` is a better name, but I guess my main point was just to update the description to explain that it's the query used when the user opens the dropdown without any filter text specified in the `<input>`.

Combobox/ComboPopup.js
wkeese: As we talked about on Sametime is actually set on line 54.

Combobox.js
wkeese: Also, _initHandler() is called every time the app changes Combobox#list, but this code to setup click and mousedown handlers should only run once.  So, move this code to createdCallback() ?

brunano21:
PR description:
This PR consists in a `Combobox` refactor, with the introduction of few enhancements.
- The widget is designed to be friendly for server-side stores.
- The widget now accepts a source. This will be assigned to the list when required.
- The widget exposes a `defaultQuery` property. This can be a function or an object and consists in the default query (`!== {}`).
- The widget can have an initial value, accessible through `widget.value`. It can be an array or a string. See `samples/Combobox.html for few examples. 
- The widget can behave as an auto-complete widget (set through hasDownArrow=false). In this scenario, the user can type some text and matching items will be shown. The auto-complete version works in mobile too, showing a popup with the a title and the input element. This mode works only when `autoFilter=true` and `selectionMode=single`.
- Add minFilterChars property. It defines the minimum number of characters the user has to type before a filtering operation can run.
- Add filterDelay property. It defines the amount of time (from the last key pressed) has to pass before a filtering operation gets executed.
- Bugfixes.

wkeese:
Replaced by #667.

PR #657, opened 2016-10-08
Button: add aria-label on the root node.

brunano21:


Button/Button.html
```diff
@@ -1,4 +1,4 @@
-<template attach-point="focusNode">
+<template attach-point="focusNode" aria-label="{{!this.showLabel ? this.label : ''}}">
```
wkeese: Actually, thinking about this more, I don't think setting `aria-label` to `""` is safe, because `aria-label` overrides the inner text of the `<button>`.  So you can end up with a button like `<button aria-label="">Submit</button>` and then a screen reader may announce that the `<button>` has no label (i.e. a blank label).

In other words `<button aria-label="">Submit</button>` is different than `<button>Submit</button>`.

brunano21:
Ok I have updated the code via 9d6afd5e77feb284e445ec4ceb14882564a3945f
<https://github.com/brunano21/deliteful/commit/9d6afd5e77feb284e445ec4ceb14882564a3945f>

Bruno Bruzzano / Software Engineer
brunano21@gmail.com / +353 83 321 9468

IBM Ireland Product Distribution Limited
brunobru@ie.ibm.com / +353 1 373 99019
IBM Technology Campus - Office B3-G-114 / Mulhuddart, Dublin 15 / Ireland

[image: Facebook]  <https://www.facebook.com/mbruno.bruzzano>[image: Google
Plus]  <https://plus.google.com/+BrunoBruzzano>[image: Youtube]
<https://www.youtube.com/channel/UCYXcigBgVM5N2ZYn8HskQew>[image: Linkedin]
<http://ie.linkedin.com/pub/bruno-bruzzano/68/b2a/6a7/it>[image: Instagram]
<http://instagram.com/_br1_/>[image: Github]
<https://github.com/brunano21>[image:
Skype]

wkeese:
Pushed as 0fd7263c24dadeb37d6d1340cb9036d6c588a6df with a few more simplifications (mainly that in a template `{{label}}` means the same thing as `{{this.label}}`).

PR #659, opened 2016-10-12
Fixes #658.

### LINKED ISSUE #658: Combobox: accessibility violation about aria-activedescendant
When the popup is open, the Combobox's `input` gets the `aria-activedescendant` attribute. 
Running an RPT scan a violation gets raised, reporting:
 
    The d-combobox element with WAI-ARIA property aria-activedescendant
    set to d-combobox-item-1 does not have tabindex attribute set to 0 or -1.

However setting the `tabindex` attribute to the `input` element is not enough yet. In order to fix the violation both `d-combobox` and `d-combobox-input` node must have the `tabindex` attribute set.

brunano21:
Fixes #658.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -738,6 +738,9 @@ define([
 				var input = this._popupInput || this.inputNode;
 				input.removeAttribute("aria-activedescendant");
 
+				// remove tabindex add into _setActiveDescendant()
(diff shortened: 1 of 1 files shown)


wkeese:
As I wrote in #658, closing this for now. The error RPT is reporting may just be a problem with RPT.

PR #660, opened 2016-10-12
Toggle button: add aria-label.

brunano21:
Fixes accessibility violations in case of **icon-only** toggle buttons.


wkeese:
Thanks, pushed as bbe0bdd189a967381d6c33188467b728fcd691b6.

Don't forget to reference a bug number in commit messages [if you are fixing a bug or doing an enhancement].

PR #667, opened 2016-11-08
Combobox refactor (new branch)

wkeese:
PR description:
This PR consists in a `Combobox` refactor, with the introduction of few enhancements.
- The widget is designed to be friendly for server-side stores.
- The widget now accepts a source. This will be assigned to the list when required.
- The widget exposes a `defaultQuery` property. This can be a function or an object and consists in the default query (`!== {}`).
- The widget can have an initial value, accessible through `widget.value`. It can be an array or a string. See `samples/Combobox.html for few examples. 
- The widget can behave as an auto-complete widget (set through hasDownArrow=false). In this scenario, the user can type some text and matching items will be shown. The auto-complete version works in mobile too, showing a popup with the a title and the input element. This mode works only when `autoFilter=true` and `selectionMode=single`.
- Add minFilterChars property. It defines the minimum number of characters the user has to type before a filtering operation can run.
- Add filterDelay property. It defines the amount of time (from the last key pressed) has to pass before a filtering operation gets executed.
- Bugfixes.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -111,8 +111,7 @@ define([
 	 *   "deliteful/Combobox", ..., "requirejs-domready/domReady!"],
 	 *   function (List, Combobox, ...) {
 	 *     var dataStore = ...; // Create data store
-	 *     var list = new List({source: dataStore});
-	 *     var combobox = new Combobox({list: list, selectionMode: "multiple"}).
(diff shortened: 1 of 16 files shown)


wkeese:
Pushed as 785de8437cb0b1d2754678a01c558fc6637ddad8 and also 314ec74400300bf1f0d06ac497bf999b00dce651.

Issue #672, opened 2016-11-22
Accordion: navigation within the header is broken.


brunano21:
It seems that keyboard navigation (tab) within a custom accordion header is broken. For this reason tests are failing.

wkeese:
Regression fixed by ibm-js/delite@e232bc563e5578b79577532921cf4ee1bf04c0ce.

Issue #671, opened 2016-11-22
Accordion test failures


brunano21:
After changes introduces by a156e55, accordion tests started to fail. That's because Accordion header's `aria-controls` attribute is set only one the accordion panel is visible. If not, it's removed indeed. 

However Accordion tests still make a deep use of the aforementioned attribute, bringing tests to fails.

wkeese:
Thanks, I pushed your fixes as fb0c1eb5c89bb165662a1c8a6c50faa8f0f03083 and 0b614c689e613e32839708cd0570ac2ee09a9103.

PR #673, opened 2016-11-22
Accordion: multiple fixes.

brunano21:
This PR description:
- It fixes failing tests, caused by changes introduced by a156e55 (#671).
- It fixes keyboard navigation within a custom Accordion header (#672).

### DIFF
--- a/Accordion/AccordionHeader.js
+++ b/Accordion/AccordionHeader.js
@@ -2,7 +2,8 @@
 define([
 	"delite/register",
 	"delite/Widget",
-	"delite/handlebars!./AccordionHeader/AccordionHeader.html"
+	"delite/handlebars!./AccordionHeader/AccordionHeader.html",
+	"delite/activationTracker"
 ], function (
 	register,
 	Widget,
--- a/tests/functional/Accordion.js
+++ b/tests/functional/Accordion.js
@@ -7,6 +7,8 @@ define(["intern",
 ], function (intern, registerSuite, pollUntil, keys, assert, require) {
 	var PAGE = "./Accordion.html";
 
+	var panelHeaderStr = "_panelHeader";
+
 	function checkPanelIsOpen(remote, panel) {
 		return remote
 			.findById(panel)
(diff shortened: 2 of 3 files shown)


brunano21:
Regarding commit #e16235c, I see that the navigation within a custom Accordion header is not the only thing broken. In fact, List is affected by the same problem, because the function at this [line](https://github.com/ibm-js/deliteful/blob/master/list/List.js#L1035) is never called, since KeyNav does not fire `delite-deactivated` event anymore. Digging more, I see that recent [KeyNav changes](https://github.com/ibm-js/delite/commit/3b7c96f074e898269f6a08f23c45c5c3329189b8) may be the reason of all these problems. 

@wkeese what do you think about?

wkeese:
I see the navigation within a custom Accordion header is working on Chrome, but not on IE11.

On IE11, when the AccordionHeader is focused, there's a focus event on the AccordionHeader (as expected) so it gets a `delite-activated` event.   So far so good.  But then there's immediately another focus event on the Accordion itself.   Thus the AccordionHeader gets a `delite-deactivated` event, and the `<button>` inside the AccordionHeader is set to tabIndex -1, so it's not tab navigable.

I'm not sure yet why the Accordion gets a focus event.  `Accordion#hasAttribute("tabindex")` is false.

wkeese:
OK, the problem is the order methods are called:

1. User tabs to Accordion.   This will cause a `focusin` event and then a `focus` event.
2. Accordion `focusin` event triggers KeyNav#focusinHandler().
4. KeyNav#focusinHandler() calls KeyNav#focus() which eventually calls AccordionHeader#focus().
5. AccordionHeader gets `focus` event which makes activationTracker send `delite-activated` event to AccordionHeader.
6. Accordion `focusin` event processing finally finished.
7 Accordion `focus` event processing starts.    That makes activationTracker send `delite-deactivated` event to AccordionHeader because it thinks focus has shifted back to the Accordion itself.

TL;DR activationTracker gets confused when focus changes inside a `focusin` handler.

I will work on a fix for that.

wkeese:
I pushed these changes after updating the checkin comments to be more descriptive, and also pushed a delite fix for the problem listed above (ibm-js/delite@ e232bc563e5578b79577532921cf4ee1bf04c0ce).

PR #674, opened 2016-11-28
Combobox: multiple fixes.

### LINKED ISSUE #677: ComboPopup: cancelHandler does not clear selected items, if any.
Valid only for Combobox's `selectionMode = multiple`
Whenever the user opens up the dropdown and selected some items, if he hits the **Cancel** button, the selection made is not cleared.

Potential fix would be to clear reset the `combo.list.selectedItems` to an empty array (`[]`) and then call `combo._validateMultiple(true)`, so the widget state gets update accordingly.

brunano21:
This PR:
- Fixes broken placeholder - _Actually I don't remember it was working before the Combobox refactor at _ 785de8437cb0b1d2754678a01c558fc6637ddad8.
- displayedValue prop refers to the value to display at each user interaction. If it's not provided at instantiation time, it will be retrieved from the source, using the value as key. Only valid if `source != null`
- Delay the widget initialization if the list has not been created yet.
- Restore down-arrow key original behaviour, even if the inputNode is empty. This means that if you press down arrow key, the popup will open - (Fixes accessibility issues).
- Fixes #677. _This bug has been introduced by_ 785de8437cb0b1d2754678a01c558fc6637ddad8
- Do not forward selection event to the list if dropdown is not opened.  _This bug has been there before the refactor at_ 785de8437cb0b1d2754678a01c558fc6637ddad8.
**Issue explanation**: In case of Combobox configured in "selectionMode=multiple", "Spacebar" events are sent to the list even though the dropdown is closed, resulting in a weird/ambiguous Combobox's selection behavior.  
selection

- Fixes broken readOnly property. _This bug has been there before the refactor at_ 785de8437cb0b1d2754678a01c558fc6637ddad8. 
**Issue explanation**: even if the widget is set i.e. in `autoFilter=false` OR `selectionMode=multiple`, at the creation time the widget's `inputNode` does not have the `readonly` attribute set, despite the `this.notifyCurrentValue("_inputReadOnly")` is executed. From my debugging I see that the notification about the `_inputReadOnly` property is delivered only when the user starts to interact with the widgets, i.e. when the user opens the dropdown. However, this is too late. In fact, if the user gets the focus on the widget via tabbing over the page and then he starts typing, the input is still "no-readonly" mode, allowing so the actual typing.
readonly

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -984,6 +984,7 @@ define([
 						}
 						this.dropDown.focus();
 					}
+
 				}.bind(this));
 			};
 		}),
--- a/Combobox.js
+++ b/Combobox.js
@@ -746,7 +746,7 @@ define([
 					if (this.opened) {
 						this.closeDropDown(true/*refocus*/);
 					}
-				} else if (evt.key === "Spacebar") {
+				} else if (evt.key === "Spacebar" && this.opened) {
 					// Simply forwarding the key event to List doesn't allow toggling
 					// the selection, because List's mechanism is based on the event target
 					// which here is the input element outside the List. TODO: see deliteful #500.
--- a/Combobox.js
+++ b/Combobox.js
@@ -1033,6 +1033,9 @@ define([
 			// Since List is in focus-less mode, it does not give focus to
 			// navigated items, thus the browser does not autoscroll.
 			// TODO: see deliteful #498
+			if (!item) {
+				item = this.list.navigatedDescendant;
+			}
 
 			if (!item) {
 				var selectedItems = this.list.selectedItems;
--- a/Combobox.js
+++ b/Combobox.js
@@ -539,6 +539,8 @@ define([
 								}
 							}.bind(this), 100); // worth exposing a property for the delay?
 						}
+					} else {
+						this.focusNode.focus(); // put the focus back to the inputNode.
 					}
 				}.bind(this)),
 
--- a/Combobox.js
+++ b/Combobox.js
@@ -669,8 +669,11 @@ define([
 				// inputNode does not contain text
 				if (!this.hasDownArrow) {
 					// in auto complete mode
-					this.closeDropDown();
-					this._toggleComboPopupList();
+					if (this._isMobile) {
+						this._toggleComboPopupList();
+					} else {
+						this.closeDropDown();
+					}
 					return false;
 				}
 			}
--- a/Combobox.js
+++ b/Combobox.js
@@ -204,7 +204,10 @@ define([
 		list: null,
 
 		// Flag used for binding the readonly attribute of the input element in the template
-		_inputReadOnly: true,
+		_inputReadOnly: false,
+
+		// Flag used to track when the inputNode has to be updated in order to set the readonly attribute.
+		_updateInputReadOnly: false,
 
 		/**
 		 * The value of the placeholder attribute of the input element used
@@ -295,6 +298,8 @@ define([
 		// Flag used for post initializing the widget value, if the list has not been created yet.
 		_processValueAfterListInit: false,
 
+		_isMobile: !!ComboPopup,
+
 		createdCallback: function () {
 			// Declarative case (list specified declaratively inside the declarative Combobox)
 			var list = this.querySelector("d-list");
@@ -312,14 +317,14 @@ define([
 
 			this.on("click", function () {
 				// NOTE: This runs only when in mobile mode
-				if (this._useCenteredDropDown() && !this.disabled) {
+				if (this._isMobile && !this.disabled) {
 					this.openDropDown();
 				}
 			}.bind(this));
 
 			this.on("mousedown", function (evt) {
 				// NOTE: This runs only when in desktop mode
-				if (!this._useCenteredDropDown() && (!this.minFilterChars || this._inputReadOnly)) {
+				if (!this._isMobile && (!this.minFilterChars || this._inputReadOnly)) {
 					// event could be triggered by the down arrow element. If so, we do not react to it.
 					if (evt.srcElement !== this.buttonNode && !this.disabled) {
 						if (!this.opened) {
@@ -381,25 +386,39 @@ define([
 
 		/* jshint maxcomplexity: 17 */
 		refreshRendering: function (oldValues) {
-			var updateReadOnly = false;
 			if ("list" in oldValues && this.list) {
 				this._initList();
 			}
 			if ("selectionMode" in oldValues) {
-				updateReadOnly = true;
 				if (this.list) {
 					this.list.selectionMode = this.selectionMode === "single" ?
 						"radio" : "multiple";
 				}
 			}
-			if ("autoFilter" in oldValues ||
-				"readOnly" in oldValues) {
-				updateReadOnly = true;
+			if ("autoFilter" in oldValues || "selectionMode" in oldValues ||
+				"readOnly" in oldValues || "source" in oldValues || "_updateInputReadOnly" in oldValues) {
+				this._updateInputReadOnly = false;
+
+				this._inputReadOnly = this.hasDownArrow ? this._isSourceEmpty() ||
+					this.readOnly || !this.autoFilter || this.selectionMode === "multiple" : false;
+
+				this.inputNode.readOnly = this._inputReadOnly || this._isMobile ? "readonly" : "";
+
+				if (this._inputReadOnly) {
+					this.inputNode.setAttribute("unselectable", "on");
+					$(this.inputNode)
+						.css("user-select", "none") // maps to WebkitUserSelect, etc.
+						.on("selectstart", false);
+				} else {
+					this.inputNode.removeAttribute("unselectable");
+					$(this.inputNode)
+						.css("user-select", "") // maps to WebkitUserSelect, etc.
+						.off("selectstart", false);
+				}
 			}
-			if (updateReadOnly) {
-				this._updateInputReadOnly();
-				this._setSelectable(this.inputNode, !this.inputNode.readOnly);
-			}
+		},
+
+		computeProperties: function (oldValues) {
 			if ("value" in oldValues) {
 				if (!this._justValidated) {
 					if (this.list) {
@@ -415,48 +434,6 @@ define([
 				if (this._justValidated) {
 					this._justValidated = false;
 				}
-			}
-			if ("displayedValue" in oldValues) {
-				this.inputNode.value = this.displayedValue;
-			}
-		},
-
-		/**
-		 * Updates the value of the private property on which the Combobox template
-		 * binds the `readonly` attribute of the input element.
-		 * @private
-		 */
-		_updateInputReadOnly: function () {
-			var oldValue = this._inputReadOnly;
-			this._inputReadOnly = this.readOnly || !this.autoFilter ||
-				this._useCenteredDropDown() || this.selectionMode === "multiple";
-			if (this._inputReadOnly === oldValue) {
-				// FormValueWidget.refreshRendering() mirrors the value of widget.readOnly
-				// to focusNode.readOnly, thus competing with the binding of the readOnly
-				// attribute of the input node (which is also the focusNode attach point)
-				// in the template of Combobox. To ensure the refresh of the binding is done
-				// including when the value of the flag _inputReadOnly doesn't change while
-				// FormValueWidget has reset the attribute to a different value, force
-				// the notification:
-				this.notifyCurrentValue("_inputReadOnly");
-			} // else no need to notify "by hand", rely on automatic notification
-		},
-
-		/**
-		 * Configures inputNode such that the text is selectable or unselectable.
-		 * @private
-		 */
-		_setSelectable: function (inputNode, selectable) {
-			if (selectable) {
-				inputNode.removeAttribute("unselectable");
-				$(inputNode)
-					.css("user-select", "") // maps to WebkitUserSelect, etc.
-					.off("selectstart", false);
-			} else {
-				inputNode.setAttribute("unselectable", "on");
-				$(inputNode)
-					.css("user-select", "none") // maps to WebkitUserSelect, etc.
-					.on("selectstart", false);
 			}
 		},
 
@@ -550,6 +527,14 @@ define([
 					this.handleOnInput(this.value); // emit "input" event
 				}.bind(this)),
 
+				this.list.observe(function (oldValues) {
+					if ("renderItems" in oldValues) {
+						// if the source changes, we may have to update the
+						// inputNode's readOnly attribute
+						this._updateInputReadOnly = true;
+					}
+				}.bind(this)),
+
 				this.list.on("query-success", this._setSelectedItems.bind(this))
 			];
 		},
@@ -596,19 +581,16 @@ define([
 		 * on the channel has() features set by `deliteful/features`.
 		 * @private
 		 */
-		_useCenteredDropDown: function () {
-			return !!ComboPopup;
-		},
+		// _useCenteredDropDown: function () {
+		// 	return !!ComboPopup;
+		// },
 
 		loadDropDown: function () {
-			this._updateInputReadOnly();
-
-			var centeredDropDown = this._useCenteredDropDown();
-			var dropDown = centeredDropDown ?
+			var dropDown = this._isMobile ?
 				this.createCenteredDropDown() :
 				this.createAboveBelowDropDown();
 
-			this.dropDownPosition = centeredDropDown ?
+			this.dropDownPosition = this._isMobile ?
 				["center"] :
 				["below", "above"]; // this is the default
 
@@ -658,12 +640,13 @@ define([
 		 * @return {boolean}
 		 */
 		shouldRunQuery: function (inputElement) {
-			var mobile = this._useCenteredDropDown();
 			if (inputElement.value.length !== 0) {
 				// inputNode contains text
 				if (inputElement.value.length < this.minFilterChars) {
-					if (!mobile) {
+					if (!this._isMobile) {
 						this.closeDropDown();
+					} else {
+						this._toggleComboPopupList();
 					}
 					return false;
 				}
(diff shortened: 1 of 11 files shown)


Combobox.js
```diff
@@ -395,11 +402,23 @@ define([
...
 				}
 			}
+			if ("_justValidated" in oldValues) {
+				if (this._justValidated) {
+					this._justValidated = false;
+				}
+			}
+			if ("displayedValue" in oldValues) {
+				this.inputNode.value = this.displayedValue;
+			}
```
wkeese: How about just saying `value={{displayedValue}}` in the template?

Combobox.js
```diff
@@ -395,11 +402,23 @@ define([
...
+						// INFO: otherwise we have to delay its evaluation.
+						this._processValueAfterListInit = true;
+					}
 				}
 			}
+			if ("_justValidated" in oldValues) {
+				if (this._justValidated) {
+					this._justValidated = false;
+				}
+			}
```
wkeese: This is confusing to me, why don't you just do:

```
this._justValidated = false;
```

all the time, without the `if()` statements?

Combobox.js
```diff
@@ -848,6 +873,21 @@ define([
...
+				if (Array.isArray(_source)) {
+					item = _source.filter(function (i) {
+						return i[this.list.valueAttr || this.list.labelAttr] === key;
+					}.bind(this))[0];
+				} else {
+					item = _source.getSync && _source.getSync(key);
+				}
+			}
+			return item;
+		},
```
wkeese: It's strange to have code that only works with a sync store, especially considering that you are working on auto-complete which is asynchronous.

Combobox.js
```diff
@@ -960,8 +1000,10 @@ define([
 		_dropDownKeyUpHandler: dcl.superCall(function (sup) {
 			return function () {
 				if (this.hasDownArrow) {
-					if (this.inputNode.value.length === 0 && this.minFilterChars === 0) {
-						this.list.query = this._getDefaultQuery();
+					if (this.inputNode.value.length === 0) {
+						if (this.minFilterChars === 0) {
+							this.list.query = this._getDefaultQuery();
+						}
```
wkeese: I guess this is the code change to "Restore down-arrow key original behaviour, even if the inputNode is empty."?

It doesn't seem quite right.  For example:

* What if the list hasn't been created yet?  Didn't we agree to always set the list's query at the same time as we create the list?
* What if the user typed a different key than the down arrow, such as backspace?

Combobox.js
```diff
@@ -295,6 +295,8 @@ define([
 		// Flag used for post initializing the widget value, if the list has not been created yet.
 		_processValueAfterListInit: false,
 
+		_isMobile: false,
```
wkeese: I think you can just say:

```
_isMobile: !!ComboPopup,
```

Combobox.js
brunano21: I think I can make this part specific for only when in not auto-complete mode (hasDownArrow=false). Do you agree?

Combobox.js
brunano21: - What if the list hasn't been created yet? Didn't we agree to always set the list's query at the same time as we create the list? 
-- Sure, I should check if the list is already created. However `this.list.query` and `this.list.source` are set at the same time, because the `_dropDownKeyUpHandler` function calls its `sup` function which calls `openDropDown` (where this.list.source gets assigned. So, what's wrong?

- What if the user typed a different key than the down arrow, such as backspace?
-- what's your point? In this case the above method (`_dropDownKeyUpHandler`) runs only when the user press the Down Arrow Key (see: https://github.com/ibm-js/delite/blob/master/HasDropDown.js#L400-L412). So, what's the problem with backspace?

brunano21:
Regarding the `_updateInputReadOnly()`, the problem seems to be serious. I already tried a solution similar the one you proposed, a while ago but it did not work. I now tried the exactly what you said, that is to update the `_inputReadOnly` property within `computeProperties()` and in `refreshRendering()` I do `setAttribute("unselectable", "on")` etc. stuff. Note: `_inputReadOnly` property is mapped into the Combobox.html template. However, even if `_inputReadOnly` is true (inspecting the widget) the `inputNode` does not have the `readonly` attribute set.

Combobox/ComboPopup.js
```diff
@@ -76,7 +76,11 @@ define([
 		 * @protected
 		 */
 		cancelHandler: function () {
+			// INFO: resetting any selected items.
+			this.combobox.list.selectedItems = [];
 			this.combobox.closeDropDown();
+			// cont: then ask to validate, so widget's value and inputNode get updated as well.
+			this.combobox._validateMultiple(true);
```
wkeese: Won't this code call _validateMultiple() even for single-selection Comboboxes?

tests/functional/ComboPopup.js
```diff
@@ -570,6 +690,21 @@ define([
...
+			if (remote.environmentType.browserName === "internet explorer") {
+				// https://github.com/theintern/leadfoot/issues/17
+				return this.skip("click() doesn't generate mousedown/mouseup, so popup won't open");
+			}
+			if (remote.environmentType.platformName === "iOS") {
+				// https://github.com/theintern/leadfoot/issues/61
+				return this.skip("click() doesn't generate touchstart/touchend, so popup won't open");
+			}
+
+			return checkMultiSelectionCancel(remote, "combo3");
```
wkeese: Since you're only using the `checkMultiSelectionCancel()` function once, maybe you should just inline the code into this test.

tests/functional/ComboPopup.js
```diff
@@ -403,7 +403,127 @@ define([
...
+						widgetValueAtLatestInputEvent: undefined, // never received
+						valueNodeValueAtLatestInputEvent: undefined,
+						widgetValueAtLatestChangeEvent: undefined,
+						valueNodeValueAtLatestChangeEvent: undefined
+					}, "after click on root node");
+			})
+			.end()
+			.findById(comboId + "_item1") // "Germany"
+			.click()
+			.sleep(500) // wait for popup to close
```
wkeese: This comment seems wrong, the ComboPopup doesn't close until you click the OK or cancel buttons.

wkeese: Actually, now I see that `checkMultiSelection()` and `checkMultiSelectionCancel()` are basically the same code, except for the last section where `checkMultiSelection()` clicks OK and `checkMultiSelectionCancel()` clicks Cancel.    So instead of having all that copy-and-pasted code can you just make a single method?

You could make `checkMultiSelection()` take a parameter (or parameters) to control whether it clicks the OK button or the cancel button.   Or alternately you could move the code that's specific to each test out of the shared function and into the tests themselves.

Combobox.js
```diff
@@ -669,8 +669,11 @@ define([
 				// inputNode does not contain text
 				if (!this.hasDownArrow) {
 					// in auto complete mode
-					this.closeDropDown();
-					this._toggleComboPopupList();
+					if (this._isMobile) {
+						this._toggleComboPopupList();
+					} else {
+						this.closeDropDown();
+					}
```
wkeese: You should have a test and a ticket for the bug you are fixing.  I also don't understand the commit comment.  It says:

> when `inputNode` was empty and in mobile, the popup used to close by itself

That makes it sound like ComboPopup would close immediately after opening, since when the ComboPopup opens the `inputNode` is empty.  Actually, I'm not sure if inputNode is the `<input>` in the original ComboBox or the `<input>` in the ComboPopup.

Combobox.js
```diff
@@ -684,8 +684,9 @@ define([
 		 * Toggles the list's visibility when ComboPopup is used (so in mobile)
 		 */
 		_toggleComboPopupList: function () {
-			if (this._useCenteredDropDown()) {
-				this.list.setAttribute("d-shown", "" + this.inputNode.value.length >= this.minFilterChars);
+			if (this._isMobile) {
+				this.list.setAttribute("d-shown",
+					"" + this.dropDown.inputNode.value.length >= this.minFilterChars);
```
wkeese: I don't see anywhere that you're setting `this._isMobile`.  I think I saw it in an earlier version of this PR, but I don't see it now, and I definitely don't see it in this commit.

Could you create an initial commit with just that refactor (changing _useCenteredDropDown() to _isMobile) and then I'll push that first and then move on to your bug fix commits?

tests/functional/ComboPopup.js
```diff
@@ -585,50 +597,50 @@ define([
...
+		// 		// https://github.com/theintern/leadfoot/issues/17
+		// 		return this.skip("click() doesn't generate mousedown/mouseup, so popup won't open");
+		// 	}
+		// 	if (remote.environmentType.platformName === "iOS") {
+		// 		// https://github.com/theintern/leadfoot/issues/61
+		// 		return this.skip("click() doesn't generate touchstart/touchend, so popup won't open");
+		// 	}
+
+		// 	return checkListInPopup(remote, "combo3", false, true);
+		// },
```
wkeese: Huh?   You just commented out a bunch of tests.

Combobox.js
```diff
@@ -746,7 +746,7 @@ define([
 					if (this.opened) {
 						this.closeDropDown(true/*refocus*/);
 					}
-				} else if (evt.key === "Spacebar") {
+				} else if (evt.key === "Spacebar" && this.opened) {
```
wkeese: Thanks, I pushed this as 7d08bcced21f680659d48dde31baa1207e3828a7.

Combobox.js
```diff
@@ -539,6 +539,8 @@ define([
 								}
 							}.bind(this), 100); // worth exposing a property for the delay?
 						}
+					} else {
+						this.focusNode.focus(); // put the focus back to the inputNode.
```
wkeese: This change should be tested (that only passes after your change to Combobox.js).  Probably just updating one of the existing tests to check focus after clicking a list item.

Actually, I loaded Combobox-decl.html without your changes, and tried both the first and the second Combobox on that page, and it seems like clicking a list's item already refocuses the Combobox if autoFilter===false, or the Combobox's `<input>` if autoFilter===true??

Also, I'm confused about what happens when:

1) IIRC Combobox has a mode where there's no `<input>` visible.  What happens then?
2) Multi-select mode.   In that case don't you want to keep focus on the list item?
3) ComboPopup.   Does this code run in the ComboPopup case, and if so, doesn't focus go to the ComboPopup's `<input>` or the original ComboBox `<input>`?   And what if the ComboPopup isn't showing an `<input>`, like the first example in ComboPopup.html?

tests/functional/Combobox.js
```diff
@@ -857,6 +857,151 @@ define([
...
+					"navigatedDescendant after two ARROW_DOWN: " + navigatedDescendant);
+			})
+			.pressKeys(keys.ARROW_DOWN)
+			.pressKeys(keys.ARROW_DOWN)
+			.sleep(500)
+			.execute(executeExpr)
+			.then(function (navigatedDescendant) {
+				assert(/^Canada/.test(navigatedDescendant),
+					"navigatedDescendant after other two ARROW_DOWN: " + navigatedDescendant);
+			})
```
wkeese: I'm confused.  The user clicks "UK" to select it, which causes the dropdown to close, and then the user re-opens the dropdown... shouldn't the focus be on "UK"?

On another note, this test keeps checking `this.list.navigatedDescendant`, but especially since this is a functional test, it seems like what you should really be testing for is where the focus goes, or for aria attributes like aria-selected, or something else that directly affects the user.

Combobox.js
wkeese: PS: Just noticed the rest of the _isMobile refactor is in your later commit titled "Fix broken readOnly (at creation time). Add tests."

Anyway, all the _isMobile changes should be consolidated into a single commit without any other changes.

Combobox.js
```diff
@@ -984,6 +984,7 @@ define([
 						}
 						this.dropDown.focus();
 					}
+
```
wkeese: This shouldn't be there.

Combobox/ComboPopup.js
```diff
@@ -58,6 +58,11 @@ define([
 							.removeClass("d-hidden");
 					}
 					this.combobox._prepareInput(this.inputNode);
+					this.combobox.observe(function (oldValues) {
+						if ("opened" in oldValues) {
+							this.inputNode.setAttribute("aria-expanded", this.combobox.opened);
+						}
+					}.bind(this));
```
wkeese: I'm confused, isn't this.combobox.opened always true whenever ComboPopup is shown?

Also (minor thing), the checkin comment since aria-described but the code change is for aria-labelledby.

tests/unit/Combobox.js
```diff
@@ -612,6 +612,43 @@ define([
...
+				combo.closeDropDown();
+				setTimeout(d.callback(function () {
+					assert.strictEqual(combo.inputNode.getAttribute("aria-expanded"), "false",
+					"aria-expanded - popup is closed: " + combo.id);
+				}), 200);
+			}));
+
+			return d;
+		},
+
```
wkeese: × internet explorer 11 on Windows 8.1 - unit tests - deliteful/Combobox: programatic - inputNode: aria-expanded attribute on open/close popup (0.046s)
AssertionError: aria-expanded - popup is open: HasDropDown_233: expected 'false' to equal 'true'
No stack or location

Combobox/ComboPopup.js
brunano21: Sorry, but I don't understand what you are referring to. `cancelHandler` function runs only when user press "Cancel" button, which is visible only if `selectionMode=multiple`.
In `selectionMode=single` we don't have any button shown.

tests/functional/ComboPopup.js
brunano21: About your first comment: yes, it was a copy-paste mistake. 
About your second comment: agree with you. A passed parameter could be enough to split the code and to keep the function still clear.

tests/functional/Combobox.js
brunano21: `checkNavigatedDescendantWithSelection` takes `combobox3` as widget to test, which is a multiple-choice combobox.
Hence, when user clicks on _UK_ item the popup does not close. 

Regarding the `navigatedDescendant`, yes we should be checking something else (too). i.e. `aria-activedescendant` which at each user interaction is set to the highlighted/focused element of the combobox's list. 
However it may be worth to keep the test on the navigatedDescendant as this ensures that changes into `delite/KeyNav` and/or `deliteful/List` do not break anything.

tests/unit/Combobox.js
brunano21: Does this happens only in IE11? 
IIRC, there were some issues with events and IE, so some other tests were disabled for this environment.

Combobox/ComboPopup.js
wkeese: Oh OK I didn't realize that.

Combobox/ComboPopup.js
wkeese: `aria-expanded` should indicate whether or not the list is displayed, which (on the master branch) is controlled by `Combobox#_togglePopupList()`, or you can get the info from `this.combobox.list.getAttribute("d-shown")`.

Usually the list is displayed, but sometimes it isn't, like the second to last test case in ComboPopup.html where hasDownArrow=false minFilterChars=3.  The list is only displayed after typing 3 characters.

Combobox/ComboPopup.html
```diff
@@ -1,8 +1,9 @@
 <template role="presentation"
 	requires="deliteful/LinearLayout, deliteful/Button">
 	<d-linear-layout style="height: 100%">
-        <div class="d-combo-popup-header">{{header}}</div>
+        <div class="d-combo-popup-header" id="{{_tag}}-{{widgetId}}-header">{{header}}</div>
 		<input d-hidden="{{this.combobox._inputReadOnly}}"
+			aria-labelledby="{{_tag}}-{{widgetId}}-header"
```
wkeese: Note: This confused me at first since tests/functional/ComboPopup.js already checks for a label for the input:

```
var checkListInPopup = function (remote, comboId, hasFilterInput, isMultiSelect) {
...
		.findByCssSelector("label[for='" + comboId + "-input']")
		.getVisibleText()
		.then(function (value) {
			assert.strictEqual(value, label);
			label = null;
		})
```

It turns out that test is useless though because it's checking the `<label>` for the `<input>` in the original Combobox rather than for  the `<input>` in the ComboPopup.

So, I see why you added this.  It's unusual to use aria-labelledby rather than `<label>` though.

wkeese: PS: Oh, I just realized that it isn't checking the `<label>` for the `<input>` in the original Combobox but rather just getting the value, to be used later.

Combobox/ComboPopup.js
wkeese: Filed #687.

Combobox/ComboPopup.html
wkeese: Filed #686.

Combobox.js
```diff
@@ -684,7 +706,7 @@ define([
 				// save what user typed at each keystroke.
 				this.value = inputElement.value;
 				if (this._useCenteredDropDown()) {
-					this.inputNode.value = inputElement.value;
+					this.displayedValue = inputElement.value;
```
wkeese: I filed #688 for the remaining task to keep displayedValue in sync w/the actual displayed value.

wkeese:
I finally got this all pushed, except for the one change we decided not to do.

Commits are efc19c57d1a7b55867de72173b1c4f7c1f779b3c, ad9a03175d2d8eaf0809c7917d09ef2e1195da6d, 5de89a363fc2b6aa1796d8a3091a88f282d138c6, 6f5417d9bd549bcf522172bac528a2737abed80f, 3196136d0cc069227ad659dab001e7ea1c05739a, 70f21429bd523eed07ce21c5d1b9dd706c96eb0a, 707d1aca28bdbb55c6f668a8cf205f4df79935dd, 11bc7ffd92e55a798eeb72fea8f4e55d2fe1700f, d9af56b53d40c9a383ea5d3e44ef32d750799e52, 2c3dbb260f2a704b4dd92a569b43acedbc7ec1ab, 668271c5e3568ddf331b552770776c699e0f2f32, ada627588256b2c1e3f136304c284b9a19a1783d, 933f1f6d67a091fd7afb952c1ed947ff2cdaad6e, 87c92d641d0caaf1526b0db6a9079cbeb68d8658, 9cfe2a1ce0f4e8aab5c39a28307613aee89ea85c.

PR #675, opened 2016-11-29
Fix the clickable area on the Combobox arrow element.

wlepinski:


Combobox/themes/bootstrap/Combobox.less
```diff
@@ -13,11 +13,12 @@
...
-	font-size: .7em;
-	right: 1em;
-	top: 50%;
-	height: 1em;
-	margin-top: -0.6em;
+	top: 0;
+	right: 0;
+	width: 28px;
+	text-align: center;
+	height: 28px;
```
wkeese: I think the current `1em` height setting is so the arrow height matches the text height (in the left part of the Combobox).   Why hardcode it to 28px?

wlepinski: Using 1em gives me around 10px because of the font-size .7em to fix the arrow size. I can calculate how much ems to use the same measurement under .7em font-size.

wlepinski: > So I think the height of the <span class="d-combobox-arrow"> needs to be specified in the same way as the height of the <input>. In other words, set the height to 1em and the top and bottom padding to @padding-base-vertical.

Sounds reasonable. I'll try to make those modifications to the PR today.

wkeese:
I checked your second version briefly.  I think there's a disparity between the `<input>` which has right padding of `1.7em` whereas the arrow's width is 12px + 12px + 9.7px (and that may vary depending on font size).   IIUC the `<input>`'s right padding should be equal to the width of the arrow.

wlepinski:
@wkeese Made a new commit cleaning up things a little bit. Now the padding-right for the input is 34px. I also added a new variable named @d-combobox-arrow-width to round the 9.7px you mentioned before. 

The calculation for the padding is now `padding: @padding-base-vertical ((@padding-base-horizontal * 2) + @d-combobox-arrow-width) @padding-base-vertical @padding-base-horizontal;`

We can also export the expression `((@padding-base-horizontal * 2) + @d-combobox-arrow-width)` to a new variable scoped to the `.d-combobox-input-styles` mixin. What do you think?

wkeese:
Thanks, I squashed the commits, updated the commit comment, and pushed as 9080840ad5055fa11c1037b587fc66dde8285729.

Issue #500, opened 2015-01-29
List: allow selection using keyboard regardless of the target of the keyboard event


AdrianVasiliu:
List handles SPACE key events via `_spaceKeydownHandler` which passes the key event to List.handleSelection. Now, this method relies exclusively on the target of the key event: if evt.target is not enclosed by an item renderer, it does nothing (see https://github.com/ibm-js/deliteful/blob/0.6.0-alpha/list/List.js#L464-471). 

This is problematic in use-cases where the List navigation is controled with key events on an element outside the List, as in the use-case of List inside the popup of Combobox. Hence, List.handleSelection could rely on the `navigatedDescendant` property for determining the element to select/deselect. However, this deserves care, at least because handleSelection is overridden in PageableList where it deals also with the prev/next page loaders elements... (https://github.com/ibm-js/deliteful/blob/0.6.0-alpha/list/PageableList.js#L771-781).

AdrianVasiliu:
Originates from @wkeese comment: https://github.com/ibm-js/deliteful/pull/497#commitcomment-9489009.

wkeese:
It's still using `event.target` but I guess this isn't an issue anymore since 6cc397814f720d396d9242d6276885015f270dfb.

PR #678, opened 2016-11-30
List#handleSelection uses navigatedDescendant instead of evt.target.

brunano21:
With List & PagebleList refactor at 3338825d03ab7b66152fd7bcae9d30d9428c2fff, the concerns expressed with #500 and [#497 comment by @wkeese](https://github.com/ibm-js/deliteful/pull/497#commitcomment-9489009) should not be valid, because `PageableList` does not override `List#handleSelection` anymore. So we can use `list.navigatedDescendant` instead of `evt.target` to find out the render the (de-)selection have to be applied on.

However this brings some List's Selection.js tests to fail.

**This is still a WIP and must not be accepted until further commits.**

### DIFF
--- a/list/List.js
+++ b/list/List.js
@@ -564,10 +564,10 @@ define([
 		 * @protected
 		 */
 		handleSelection: function (/*Event*/event) {
-			var eventRenderer = this.getEnclosingRenderer(event.target);
-			if (eventRenderer) {
-				if (!this.isCategoryRenderer(eventRenderer)) {
-					this.selectFromEvent(event, eventRenderer.item, eventRenderer, true);
+			var renderer = this.getEnclosingRenderer(this.navigatedDescendant);
+			if (renderer) {
+				if (!this.isCategoryRenderer(renderer)) {
+					this.selectFromEvent(event, renderer.item, renderer, true);
 				}
 				return true;
 			}


list/List.js
```diff
@@ -564,10 +564,10 @@ define([
 		 * @protected
 		 */
 		handleSelection: function (/*Event*/event) {
-			var eventRenderer = this.getEnclosingRenderer(event.target);
-			if (eventRenderer) {
-				if (!this.isCategoryRenderer(eventRenderer)) {
-					this.selectFromEvent(event, eventRenderer.item, eventRenderer, true);
+			var renderer = this.getEnclosingRenderer(this.navigatedDescendant);
```
wkeese: Should just be able to use `this.navigatedDescendant` rather than `this.getEnclosingRenderer(this.navigatedDescendant)`, but that doesn't explain the test failures.

brunano21:
Tests are failing because of code like the one at 
https://github.com/ibm-js/deliteful/blob/master/tests/unit/list/Selection.js#L29
This test is emulating an event, so at that stage the `navigatedDescendant` is not set. 
The solution would be to change the test in order to set `navigatedDescendant`.

wkeese:
Hmm, yes, that test should definitely be updated.   There should be a better way to write the test though than setting properties like `navigatedDescendant` (which is supposed to be read only) and calling private methods like `spaceKeyHandler()`.

wkeese:
It's still using `event.target` but I guess this isn't an issue anymore since 6cc397814f720d396d9242d6276885015f270dfb.

PR #682, opened 2017-01-04
ComboPopup: Fix cancel button's handler.

### LINKED ISSUE #677: ComboPopup: cancelHandler does not clear selected items, if any.

brunano21:
Fix cancel button's handler in order to clear any selected items.
Add tests.
Fixes #677.

### DIFF
--- a/Combobox/ComboPopup.js
+++ b/Combobox/ComboPopup.js
@@ -67,7 +67,7 @@ define([
 		 * @protected
 		 */
 		okHandler: function () {
-			this.combobox._validateMultiple(this.combobox.inputNode);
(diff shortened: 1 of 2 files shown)


wkeese:
Looks great except for one jshint error.  I fixed that and pushed as ad9a03175d2d8eaf0809c7917d09ef2e1195da6d.

Issue #668, opened 2016-11-08
Combobox enhancements to be used as auto-complete widget


wkeese:
Enhancements to Combobox so it can be used as an auto-complete widget.

- The widget is designed to be friendly for server-side stores.
- The widget now accepts a source. This will be assigned to the list when required.
- The widget exposes a `defaultQuery` property. This can be a function or an object and consists in the default query (`!== {}`).
- The widget can have an initial value, accessible through `widget.value`. It can be an array or a string. See `samples/Combobox.html for few examples. 
- The widget can behave as an auto-complete widget (set through hasDownArrow=false). In this scenario, the user can type some text and matching items will be shown. The auto-complete version works in mobile too, showing a popup with the a title and the input element. This mode works only when `autoFilter=true` and `selectionMode=single`.
- Add minFilterChars property. It defines the minimum number of characters the user has to type before a filtering operation can run.
- Add filterDelay property. It defines the amount of time (from the last key pressed) has to pass before a filtering operation gets executed.
- Bugfixes.

PR #685, opened 2017-01-05
Restore down-arrow key original behaviour, even if the inputNode is empty,

brunano21:
Restore down-arrow key original behaviour, even if the inputNode is empty. This means that if you press down arrow key, the popup will open - (Fixes accessibility issues).
Refs #668.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -268,7 +268,7 @@ define([
 		source: null,
 
 		/**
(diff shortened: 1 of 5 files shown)


wkeese:
This has been fixed in 6f5417d9bd549bcf522172bac528a2737abed80f.  See also subsequent commit 3196136d0cc069227ad659dab001e7ea1c05739a.

PR #700, opened 2017-06-08
Combobox compatible with Aria 1.1.

wlepinski:
- Depends on https://github.com/ibm-js/delite/pull/483
- Disable the move procedure of aria attributes to the focusNode.
- Make setAttribute, getAttribute, hasAttribute and removeAttribute use the HTMLElement.prototype method.
- _setActiveDescendant should set the attribute on the root node accordingly to the spec.

### DIFF
--- a/Combobox.js
+++ b/Combobox.js
@@ -459,14 +459,6 @@ define([
 			}
 		},
 
-		setAttribute: HTMLElement.prototype.setAttribute,
-
-		getAttribute: HTMLElement.prototype.getAttribute,
-
-		hasAttribute: HTMLElement.prototype.hasAttribute,
-
-		removeAttribute: HTMLElement.prototype.removeAttribute,
-
 		/**
 		 * Configures inputNode such that the text is selectable or unselectable.
 		 * @private
@@ -949,7 +941,7 @@ define([
 
 				return promise.then(function () {
 					this.setAttribute("aria-owns", this.dropDown.id);
-					this.popupStateNode.setAttribute("aria-controls", this.dropDown.id);
+					this.inputNode.setAttribute("aria-controls", this.dropDown.id);
 
 					// Avoid that List gives focus to list items when navigating, which would
 					// blur the input field used for entering the filtering criteria.
@@ -974,7 +966,7 @@ define([
 				input.removeAttribute("aria-activedescendant");
 
 				this.removeAttribute("aria-owns");
-				this.popupStateNode.removeAttribute("aria-controls");
+				this.inputNode.removeAttribute("aria-controls");
 
 				// Closing the dropdown represents a commit interaction, unless the dropdown closes
 				// automatically because the user backspaced, in which case suppressChangeEvent is true.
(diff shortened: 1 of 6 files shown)


Combobox.js
```diff
@@ -1010,8 +1031,7 @@ define([
 					nd.id = "d-combobox-item-" + idCounter++;
 				}
 
-				var input = this._popupInput || this.inputNode;
-				input.setAttribute("aria-activedescendant", nd.id);
+				this.setAttribute("aria-activedescendant", nd.id);
```
wkeese: I'm confused about this change, from looking at https://www.w3.org/TR/wai-aria-1.1/#h-combobox it seems like `aria-activedescendant` should be on the `<input>`.

Combobox.js
```diff
@@ -930,6 +945,9 @@ define([
 				var promise = sup.apply(this, arguments);
 
 				return promise.then(function () {
+					this.setAttribute("aria-owns", this.dropDown.id);
+					this.popupStateNode.setAttribute("aria-controls", this.dropDown.id);
```
wkeese: We talked about making `popupStateNode` point to the root node and then all you need is to set `aria-controls` on the `<input>`.

Combobox.js
```diff
@@ -930,6 +937,9 @@ define([
 				var promise = sup.apply(this, arguments);
 
 				return promise.then(function () {
+					this.setAttribute("aria-owns", this.dropDown.id);
+					this.inputNode.setAttribute("aria-controls", this.dropDown.id);
```
wkeese: I think the `aria-owns` is now being handled by HasDropDown so all you need is to set `aria-controls`.

Combobox/Combobox.html
```diff
@@ -1,6 +1,5 @@
-<template class="d-combobox" role="presentation">
+<template class="d-combobox" role="combobox" aria-expanded="{{this.opened}}" attach-point="popupNode">
```
wkeese: You can (and should) actually just do `{{opened}}` rather than `{{this.opened}}`.

wlepinski: Done.

wkeese:
Looks good.  Does the test file pass DAP (minus complaints about the test file itself)?

wlepinski:
> Looks good. Does the test file pass DAP (minus complaints about the test file itself)?
No complaints about the Combobox itself. 🌝

wkeese:
I pulled down the code.  It passes jshint but unfortunately is getting errors in the ComboPopup test, for example:

```
× firefox 53 on Windows 7 - ComboPopup - functional - list in popup (combo2)
```

This is from `grunt test:remote` against SauceLabs.

wkeese:
Planning below for check in comment (after squashing all the commits, including the fix to the regression test):

```
   Make Combobox compatible with Aria 1.1.
    
    - Depends on ibm-js/delite#483 and ibm-js/delite#486.
    - Disable the move procedure of aria attributes to the focusNode.
    - _setActiveDescendant should set the attribute on the root node accordingly to the spec.
    - See https://www.w3.org/TR/wai-aria-1.1/#h-combobox.
```

If there's anything else that needs to be said let me know.

wkeese:
Looks good, pushed as 83266a826988e40535826b8c637ec62b93db6ea2.

PR #706, opened 2017-07-25
Moving the active-descendent attribute to inputNode

wlepinski:
I guess during the migration to aria 1.1 I did a mistake about which node should receive the attribute.


wkeese:
Thanks, pushed as e5430ca475f21fd02f597212a2bb156ea5430c2e.   Next time remember to add "Refs #..." to commit messages.

PR #709, opened 2017-08-30
Removed attributes that prevents the user to focus the buttons

### LINKED ISSUE #708: Can't focus dialog buttons on ComboPopup using iOS VoiceOver
The user is unable to select the buttons **CANCEL** and **OK** using swipe to change the currently focused element. This is due to the fact that some aria-attributes close the context where the user can change the focus.

The attributes **role**, **aria-expanded**, **aria-owns** and **aria-haspopup** aren't necessary at the root tag.

wlepinski:
These attributes were "closing" the "focus context" for users using iOS + VoiceOver.
Removing the attributes open the sibling context and user is able to focus the buttons using swipe or keyboard shortcuts.

Fix #708

### DIFF
--- a/Combobox/ComboPopup.js
+++ b/Combobox/ComboPopup.js
@@ -45,6 +45,11 @@ define([
 					(combobox.hasAttribute("aria-labelledby") &&
 					this.ownerDocument.getElementById(combobox.getAttribute("aria-labelledby")));
 				this.header = headerNode ? headerNode.textContent.trim() : (combobox.getAttribute("aria-label") || "");
+				
+				var list = this.combobox.list;
+				if (list) {
+					this.onKeynavChildNavigatedEvent = list.on("keynav-child-navigated", this._onKeynavChildNavigatedHandler, this);	
+				}
 			}
 		},
 
@@ -53,12 +58,38 @@ define([
 				if (this.combobox) {
 					var list = this.combobox.list;
 					if (list) {
+						this.inputNode.removeAttribute("aria-activedescendant");
 						list.placeAt(this.listNode, "replace");
 						$(list).addClass("fill")
 							.removeClass("d-hidden");
 					}
 					this.combobox._prepareInput(this.inputNode);
 				}
+			}
+		},
+		
+		/**
+		 * Event handler called when a event of type "keynav-child-navigated" is dispatched
+		 * on the combobox.list element.
+		 * @param Event evt The Event instance. 
+		 */
+		_onKeynavChildNavigatedHandler: function (evt) {
+			var navigatedChild = evt.newValue;
+			this._setActiveDescendant(navigatedChild);
+		},
+		
+		/**
+		 * Sets the active descendant on the inputNode.
+		 * @param String nd The select node.
+		 * @protected
+		 */
+		_setActiveDescendant: function (nd) {
+			if (nd) {
+				if (!nd.id) {
+					nd.id = "d-combobox-item-" + idCounter++;
+				}
+
+				this.inputNode.setAttribute("aria-activedescendant", nd.id);
 			}
 		},
 
@@ -68,6 +99,7 @@ define([
 		 */
 		okHandler: function () {
 			// NOTE: no need to validate since it's handled by the `selection-change` listener
+			this.onKeynavChildNavigatedEvent.remove();
 			this.combobox.closeDropDown();
 		},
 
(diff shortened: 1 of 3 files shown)


wkeese:
Looks like there's a bunch of accessibility issues that should be fixed with the combo popup, including #708 that ironically I cannot reproduce.

(1) The outer node should probably have `role=dialog`.

(2) The `<label for=...>` tag doesn't work if the `<input>` is hidden because the `for` attribute doesn't make sense.   Should probably be replaced by a `<span>` where both the root node and the `<input>` have `aria-labelledby="id-of-the-span"`.

(3) `aria-expanded` is currently set on the `<input>` but that's invalid.

(4) With the change in this PR I think you need an `aria-haspopup="listbox"` on the `<input>`.  As stated in https://www.w3.org/TR/wai-aria-1.1/#aria-autocomplete:

> If an element has aria-autocomplete set to list or both, authors must ensure both of the following conditions are met:
>
> 1. The element has a value specified for aria-controls that refers to the element that contains the collection of suggested values.
> 2. Either the element or a containing element with role combobox has a value for aria-haspopup that matches the role of the element that contains the collection of suggested values.

Finally, it's weird for me to not have a `role=combobox` at all but I guess it's not necessarily an error.   I couldn't find any examples on the web though where `aria-autocomplete` is used without `role=combobox`.  If it turns out that we do need (or want) a `role=combobox` after all, I think it's easy to implement with something like shown on https://www.w3.org/TR/wai-aria-1.1/#h-combobox, i.e:

```
<div role=dialog>
	<div aria-label="Tag" role="combobox" aria-expanded="true" aria-owns="owned_listbox" aria-haspopup="listbox">
		<input type="text" aria-autocomplete="list" aria-controls="owned_listbox" aria-activedescendant="selected_option">
	</div>
	<ul role="listbox" id="owned_listbox">
		<li role="option">Zebra</li>
		<li role="option" id="selected_option">Zoom</li>
	</ul>
	<button>OK</button>
	<button>Cancel</button>
</div>
```

In other words a `role=combobox` doesn't need to contain the buttons.  It doesn't even need to contain the `role=listbox`.

wkeese:
OK thanks, that's looking better.  A few things:

1) The dialog title `<span>` has `d-shown="{{ header !== '' }}"`, which presumably won't work because the `this.` is only optional in simple cases like `{{header}}`, not in expressions like `header !== ''`.  I'm not sure setting `d-shown=false` on an empty `<span>` does much anyway.

2) Looks like the `<input>` doesn't have `aria-labelledby` but presumably it's needed (all `<input>`s need labels).

3) `<button>` nodes shouldn't need a tabindex=0 setting as that's the default.

wkeese:
Oh also, you need to run `grunt jshint`, it's getting errors.

Combobox/ComboPopup.js
```diff
@@ -61,13 +67,39 @@ define([
...
+		 */
+		_setActiveDescendant: function (nd) {
+			if (nd) {
+				if (!nd.id) {
+					nd.id = "d-combobox-item-" + idCounter++;
+				}
+
+				this.inputNode.setAttribute("aria-activedescendant", nd.id);
+			}
+		},
```
wkeese: This code seems mainly useless since the template already has this on the `<input>`:

```
aria-activedescendant="{{combobox.list.navigatedDescendant}}"
```

The one new piece of functionality is line 89, about setting ids on the list nodes, but that code doesn't work because there's no `idCounter` variable.

tests/functional/ComboPopup.js
```diff
@@ -179,7 +179,7 @@ define([
 			// Get the ComboPopup's <input>'s id...
 			.findByCssSelector("#" + comboId + "_dropdown input").getAttribute("id").then(function (inputId) {
 				// And then get the <label> pointing to that <input>...
-				return remote.findByCssSelector("#" + comboId + "_dropdown label[for=" + inputId + "]")
+				return remote.findByCssSelector("#" + comboId + "_dropdown .d-combo-popup-header")
 					.getVisibleText().then(function (popupLabel) {
 						// And then make sure it matches the Combobox's label.
 						return remote.findByCssSelector("label[for=" + comboId + "-input]")
```
wkeese: I will check in a modified version of this change.  It turns out there's no reason to looking up the `id` of the `<input>` at all.

Combobox/ComboPopup.js
wkeese: Perhaps the code in the template isn't working because it doesn't realize when combobox.list.navigatedDescendant changes?

wkeese:
This was actually checked in a long time ago in 5947b3cbc4c7c75ee6d5846342bbb738b8c72c48.

